PR #1, opened 2017-07-25
Add file for initial API documentation

jonc125:
This will be filled in by Jonathan based on the existing fcweb code, as a basis for the Django re-write. It is not intended to be the final documentation - that should be auto-generated using something like swagger.

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -16,6 +16,8 @@ A `?` after a JSON field name indicates that the field is optional.
 
 Arguably we'd be more RESTful to turn some of these POST requests into GET ones, since they don't actually change anything on the server...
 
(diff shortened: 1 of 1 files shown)


jonc125:
I've now finished outlining the existing REST APIs served by the Java code, which you'll hopefully find helpful for the rewrite!

docs/rest_api.md
```diff
@@ -0,0 +1,114 @@
...
+The class also has further utility methods for communicating with the Celery experiment runner (sending messages), e.g. `submitExperiment`, `getProtocolInterface`, `cancelExperiments`
+
+## Index.java
+
+Just displays the index page. No API calls.
+
+## Login.java, Logout.java, Register.java
+
+Just handle user login/logout/registration, via web pages & API calls; will be obsolete.
+
```
helenst: indeed! Django will take care of this nicely. :)

Issue #5, opened 2017-08-01
Set up static files


helenst:
For now, the existing set of static files can be ported directly across and hooked into the new back end.
However, it would be helpful to make some basic improvements to make the code easier to work with:

* Static build system (e.g. gulp)
* Improve the way javascript files are loaded (e.g. bundler?)
* Set up Sass build (this can be done incrementally as css is a subset of scss)

helenst:
This was done as part of #7

PR #7, opened 2017-08-07
Basic Django project setup

### LINKED ISSUE #4: Create skeleton Django project
Create a basic Django project structure. This should include at least:

* database configuration
* requirements files
* test runner
* basic instructions for a developer to install and run
* some set up for static file location (not necessary to import all static files at this point)

helenst:
This is a simple Django project set up with a hello world page using the base template from the old project. There's a database as well as some testing setup. I've imported the static files from the old project and set up a sass build (css can be renamed to scss and built, so those files can be adapted over time to be more sass-like) and also a JS build using browserify - just the files in the base template at this point, but more can be added to that over time.

Fixes #4 and #5


jonc125:
Looks good! I've got it installed locally and managed to visit the base page. I added some further tweaks in the django-project branch in this repo, since I don't have push access to your fork. Would be good to include those before merging.

I'm also getting `ImportError: No module named 'config.settings.test'` when running pytest, but I assume that's basically because there aren't any tests yet.

helenst:
@jonc125 Thank you for the extra commits. I've brought those into my branch and also added the test config file.

helenst:
@jonc125 yes, I'm happy for this to be merged!

Issue #6, opened 2017-08-01
Set up login system


helenst:
Make sure users can:
- [x] Register
- [x] Log in
- [x] Reset password
- [x] Change password
- [x] View account details
- [x] Log in with GitHub
- [x] Log in with Google

Also:
- [x] add email flag
- [x] email admins when new user registers
- [x] icons / visuals for oauth logins: display on registration page too
- [x] edit email address if registered locally only (no oauth)
- [x] Change password could be a button
- [x] Get rid of username field and just use email address to identify

The Django system provides a lot of help with these features, but templates and URL config will still need to be provided, as well as setting up a suitable user model.

jonc125:
How easy is it to add OAuth authentication (via e.g. GitHub and/or Google)? @mirams requested this earlier today, to avoid needing another password!

helenst:
@jonc125 https://github.com/python-social-auth should help with this... it's the same thing UCL Dashboard uses (it can be configured to authenticate rather than just connect to existing account)

jonc125:
That would be nice to have as an option then, if it's not much work. How long do you think it would take?

Also, is it possible for site admins to be sent an email when a user registers?

helenst:
@jonc125 Shouldn't be a lot of work. At a guess, half a day, max 1 day.

Sending an email should be no problem.

jonc125:
@helenst Great, could you include that feature then please!

PR #9, opened 2017-08-23
User registration and login

helenst:
Still a few more things to do / tidy up here (see #6) but putting this out for comment in the meantime. This sets up a new user model with login, registration, password reset and user profile page. It also supports login via Google and GitHub.

If you are testing this out and already have a database, I recommend recreating that database from scratch, since custom user models set up migration dependencies that are awkward to deal with on an existing database - by re-running all the migrations from the start, it's a lot easier.


helenst:
@jonc125 Taking this out of WIP status so you can review. Looks like the travis stuff worked as this is now building and tests passing!

weblab/templates/emails/user_created.txt
```diff
@@ -0,0 +1,5 @@
+A new WebLab user was created.
```
jonc125: Is it possible to add in the user_created email a link to the admin page where an email can give them additional permissions? Because that's the most common thing we'll want to do when someone registers - allow them to upload models if we know them!

weblab/templates/includes/navbar.html
```diff
@@ -2,6 +2,25 @@
 <nav>
   <ul class="navbar">
     <li><a href="{% url "home" %}">Home</a></li>
+    {% if user.is_authenticated %}
+        <li id="usernavbar">
+          <a>{{ user.full_name }}</a>
+          <ul>
+            <!--li><a href="/myfiles">My files</a></li-->
```
jonc125: Should this be removed? Seems it's included below uncommented.

weblab/templates/registration/register.html
```diff
@@ -0,0 +1,21 @@
...
+
+  <p>
+    Only users with an account are able to upload their own models and/or
+    protocols for analysis. Once you have registered, you will need to wait
+    for an admin to approve the account before you are able to upload files.
+  </p>
+
+  <form action="" method="POST">
+    {% csrf_token %}
+    {{ form.as_p }}
```
jonc125: When you register in Firefox, the password details it offers to save have your institution as the username! Is it possible to tell Firefox that the email address is the username instead?

weblab/templates/registration/myaccount.html
```diff
@@ -0,0 +1,76 @@
...
+          {{ form.receive_emails.label }}
+          <span class="errorlist">{{ form.receive_emails.errors }}</span>
+        </td>
+      </tr>
+    </table>
+
+    <p>
+      <button>Update account</button>
+      {% if user.has_usable_password %}
+        <a class="button" href="{% url "password_change" %}">Change Password...</a>
```
jonc125: Interestingly, this has subtly different display from a standard button, at least on Firefox - it's a smidgen taller!

weblab/templates/registration/login.html
```diff
@@ -0,0 +1,27 @@
...
+
+  <p>
+    <a class="google-icon" href="{% url "social:begin" "google-oauth2" %}"><img src="{% static 'img/oauth-icons/google.png' %}" /></a>
+  </p>
+  <p>
+    <a class="github-icon" href="{% url "social:begin" "github" %}"><img src="{% static 'img/oauth-icons/github.png' %}" /><span>Sign in with GitHub</span></a>
+  </p>
+
+  <p>
+    <a href="{% url "password_reset" %}">Reset Password</a>
```
jonc125: Can this be made a button too?

weblab/templates/registration/register.html
helenst: Sure... unfortunately firefox (and possibly other browsers?) uses the field before the password field as username if it can't find one called username. I've changed the order to accommodate this.

helenst:
I've addressed the comments and this is ready again from my perspective!

jonc125:
Looks good to me too!

PR #34, opened 2018-01-04
Entity visibility implementation

### LINKED ISSUE #24: Visibility settings for entities
Implement visibility for viewing entities - public, restricted or private. The setting is already present on the model, but its effects need to be implemented.

helenst:
Implement entity visibility settings (public, private, restricted) for entity views

Fixes #24


jonc125:
Is this ready for review?

helenst:
@jonc125 Yes please! :)

requirements/dev.txt
```diff
@@ -44,7 +45,7 @@ pytz==2017.2              # via django
 requests-oauthlib==0.8.0  # via social-auth-core
 requests==2.18.4          # via requests-oauthlib, social-auth-core
 simplegeneric==0.8.1      # via ipython
-six==1.10.0               # via pip-tools, prompt-toolkit, social-auth-app-django, social-auth-core, traitlets
+six==1.10.0               # via model-mommy, pip-tools, prompt-toolkit, social-auth-app-django, social-auth-core, traitlets
```
MichaelClerx: Where do these `via`s come from? Do we add them manually or are they generated somehow?

jonc125: They're generated by pip-tools compiling the `dev.in` file.

weblab/core/recipes.py
```diff
@@ -0,0 +1,10 @@
+from model_mommy.recipe import Recipe, seq
+
+
+user = Recipe('accounts.User', institution='UCL')
+
+model = Recipe('ModelEntity', visibility='public', entity_type='model', name=seq('mymodel'))
```
MichaelClerx: Is this to make our testing code easier to write? Or are there performance benefits involved?

weblab/entities/models.py
```diff
@@ -61,7 +62,7 @@ def add_file_to_repo(self, file_path):
         self.repo.index.add([file_path])
 
     def commit_repo(self, message, author_name, author_email):
```
MichaelClerx: I'd really like some docstrings here! What does this method return? What happens with git or connection errors?

weblab/entities/tests/test_views.py
```diff
@@ -43,6 +49,15 @@ def fake_upload_path(settings, tmpdir):
     return settings.MEDIA_ROOT
 
 
+def add_version(entity):
+    """Add a single commit/version to an entity"""
+    entity.init_repo()
+    in_repo_path = str(entity.repo_abs_path / 'entity.txt')
```
MichaelClerx: What's happening here? Has the `/` operator overloaded?

weblab/entities/tests/test_views.py
```diff
@@ -98,22 +113,85 @@ def test_create_protocol_requires_permissions(self, user, client):
...
-            entity=model,
+        model = recipes.model_file.make(
+            entity__author=user,
             upload=SimpleUploadedFile('model.txt', b'my test model'),
             original_name='model.txt',
-        )
+        ).entity
         response = client.post(
-            '/entities/models/' + str(model.pk) + '/versions/new',
+            '/entities/models/%d/versions/new' % model.pk,
```
MichaelClerx: I thought the `%` operator was being phased out? https://www.python.org/dev/peps/pep-3101/

weblab/core/recipes.py
MichaelClerx: Not (just) being grumpy here, but a bit concerned about adding dependencies (on what looks to be a 2 person project) for simple things like object creation! Every new dependency has the potential to increase the learning curve for new developers and cause issues for installation / testing / cross-platform availability...

weblab/entities/views.py
```diff
@@ -6,6 +6,7 @@
 from django.contrib.auth.mixins import (
```
MichaelClerx: Awesome! I didn't know this syntax

weblab/entities/views.py
```diff
@@ -25,10 +26,13 @@
     ModelEntityForm,
     ProtocolEntityForm,
 )
-from .models import ModelEntity, ProtocolEntity
+from .models import Entity, ModelEntity, ProtocolEntity
 
 
 class ModelEntityTypeMixin:
+    """
+    Mixin for model-based pages
```
MichaelClerx: What does it do?

weblab/entities/tests/test_views.py
jonc125: There is a new `.format` approach, but `%` is not yet deprecated as far as I know.

weblab/entities/tests/test_views.py
jonc125: This is [pathlib](https://docs.python.org/3/library/pathlib.html), new in Python 3.4.

requirements/dev.txt
jonc125: Incidentally, it seems the new `dev.in` has been missed from the PR!

weblab/entities/views.py
```diff
@@ -91,32 +107,72 @@ def get_success_url(self):
...
+    List all user's protocol entities
+    """
     template_name = 'entities/entity_list.html'
 
     def get_queryset(self):
         return self.model.objects.filter(author=self.request.user)
 
 
-class ModelEntityVersionView(LoginRequiredMixin, ModelEntityTypeMixin, VersionMixin, DetailView):
+class EntityAccessMixin(UserPassesTestMixin):
```
MichaelClerx: I like this mix-in approach! But not sure about the "access" part of the name, as the checks concern _visibility_. Maybe use something like EntityViewingMixin or EntityVisibleMixin to emphasize it doesn't imply editing rights?

weblab/entities/tests/test_views.py
MichaelClerx: Thanks! That looks pretty cool. Wonder if anyone will backport it...

weblab/entities/views.py
helenst: That might be better, thank you.

weblab/core/recipes.py
helenst: @MichaelClerx It makes testing code easier to write and clearer to read (prevents repetition of setting up irrelevant but required fields on test objects). The dependency is contained entirely within the pip dependencies. I feel like it's useful to have more concise tests, but interested in what others think.

requirements/dev.txt
helenst: It shouldn't be needed as nothing has changed directly in the dev dependencies. `dev.in` includes `test.in` which did change, which is why the resulting `dev.txt` has changed.

weblab/entities/tests/test_views.py
helenst: I think % often is the most concise option. (until you get to Python 3.6 and start using f-strings :)

weblab/entities/views.py
helenst: I'll clarify the comment.

jonc125:
@helenst I tried applying migrations and got this error message saying the models have changes that aren't in a migration yet:
```
[:~/projects/WebLab/WebLab] [weblab] entity-visibility ± weblab/manage.py migrate
Operations to perform:
  Apply all migrations: accounts, admin, auth, contenttypes, entities, sessions, social_django
Running migrations:
  No migrations to apply.
  Your models have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

jonc125:
Oh, and creating a new protocol currently says 'Create new model' :)

requirements/dev.txt
jonc125: Ah, I see - sorry!

weblab/core/recipes.py
jonc125: It's been used (and useful) on the Dashboard project too, and I agree about the benefits of clearer test code. We'd probably end up writing such helpers ourselves if we didn't use it...

jonc125:
Also, when you view a version, its visibility isn't shown. This should probably be ported from the old site as part of this PR.

weblab/entities/tests/test_views.py
```diff
@@ -168,41 +237,106 @@ def test_create_protocol_version(self, user, client):
...
+        model = recipes.model.make(visibility='private', author=user)
+        add_version(model)
+        assert client.get('/entities/models/%s' % model.pk, follow=True).status_code == 200
+
+    def test_private_entity_invisible_to_other_user(self, client, user, other_user):
+        model = recipes.model.make(visibility='private', author=other_user)
+        add_version(model)
+        response = client.get('/entities/models/%s' % model.pk)
+        assert response.status_code == 302
+        assert '/login' in response.url
```
jonc125: While redirecting to the login page is the right thing where the user is anonymous, I'm not sure it's so logical if the user is logged in but doesn't have permissions. In that case wouldn't a 403 or 404 response be more appropriate (depending on whether we want people to know the ID exists)?

weblab/entities/tests/test_views.py
```diff
@@ -168,41 +237,106 @@ def test_create_protocol_version(self, user, client):
...
+        add_version(model)
+        assert client.get('/entities/models/%s/versions/' % model.pk).status_code == 200
+
+    def test_private_entity_invisible_to_other_user(self, client, user, other_user):
+        model = recipes.model.make(visibility='private', author=other_user)
+        add_version(model)
+        # redirect to login
+        response = client.get('/entities/models/%s/versions/' % model.pk)
+        assert response.status_code == 302
+        assert '/login' in response.url
```
jonc125: Same comment as above.

weblab/entities/views.py
jonc125: I like `EntityVisibleMixin`.

weblab/entities/models.py
helenst: I'm adding some more docstrings to these repository methods. Will aim to keep docstring coverage more up to date in future commits...

weblab/entities/tests/test_views.py
```diff
@@ -98,22 +113,85 @@ def test_create_protocol_requires_permissions(self, user, client):
         assert '/login/' in response.url
 
 
+@pytest.mark.django_db
+class TestEntityDetail:
+    def test_redirects_to_latest_version(self, client, user):
+        model = recipes.model.make()
+        add_version(model)
+        commit = next(model.commits)
+        response = client.get('/entities/models/{}'.format(model.pk, commit.hexsha))
```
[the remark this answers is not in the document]
helenst: Ah yes. I hadn't realised `format` doesn't complain about that in the same way the `%` operator does!

weblab/entities/tests/test_views.py
helenst: Good call, I agree. Anonymous user should see entity if they have access, otherwise login redirect (regardless of whether it exists). Logged in user should see entity if they have access, otherwise 404.

helenst:
OK, I think I have addressed the all the comments raised - this is ready for re-review.

helenst:
@jonc125 Ah yes, I missed creating a migration when I changed something on a model Meta. I've added this now.

jonc125:
Assuming @MichaelClerx is happy with the new docstrings, this is good to merge.

PR #37, opened 2018-01-18
Improve version/tag handling & display

### LINKED ISSUE #27: Improvements to entity upload & display
Child of #12 with minor issues that didn't need to be addressed in #21.

- [x] The 'Version' field should now be optional I think. It's basically a git tag isn't it?
- [x] So rename to 'Tag' to make that clearer.
- [x] I'm not sure I like displaying the full SHA for a version. Perhaps do this only if there isn't a tag, and just show the tag otherwise? And/or show just the first 8 characters?
- [x] I like that you can use the tag in the URL though! Add a test to check this keeps working.
- [x] The versions list should show the commit message. (Or at least, the first line. I'm not sure if or where to display a longer message.)
- [x] Make the visibility status display more like the original Web Lab
- [x] Improve hover text on new version form
- [x] When generating URLs, supply a tag instead of SHA if present?
- [x] Show all tags associated with a commit (will do this in a separate PR I think)
- [x] Allow adding a tag to a version after creation (may become separate issue/PR if non-trivial)

Thoughts @mirams @MichaelClerx ?

jonc125:
Fixes (most of!) #27


jonc125:
I think this part is ready for review now. There's also the linked PR #38 which builds on it; I don't mind whether you look at this one first then that, or merge #38 and review them both together!

weblab/entities/models.py
```diff
@@ -16,6 +16,11 @@ class Entity(models.Model):
         (VISIBILITY_RESTRICTED, 'Restricted'),
         (VISIBILITY_PUBLIC, 'Public')
     )
+    VISIBILITY_HELP = (
+        'Public = anyone can view\n'
+        'Restricted = logged in users can view\n'
```
MichaelClerx: Should this not be public/private/group or something? Not sure why logged in users are a privileged group (anyone can make an account right?)

weblab/templates/base.html
```diff
@@ -1,20 +1,8 @@
...
-
-
--->
-
 <!DOCTYPE html>
 <html>
   <head>
     <title>
-      {% block title %}{% endblock %}Functional Curation @ Chaste - Cardiac Electrophysiology
+      {% block title %}{% endblock %}Web Lab - Cardiac Electrophysiology
```
MichaelClerx: Are we using this name to indicate the other future uses? I've been calling it the "Cardiac Electrophysiology Web Lab" :-)

jonc125: That was what I was thinking; well that and making a smaller edit :)

weblab/entities/models.py
jonc125: Good point. We used to have restricted when it was harder to make an account; with #40 it probably becomes redundant. I'll make a note there.

PR #38, opened 2018-01-18
Better support for multiple tags

jonc125:
This will contain the two larger items from #27.


jonc125:
Hi @helenst. Would you mind giving this and #37 a look over to see if I'm doing anything silly from a Django perspective?

weblab/templates/entities/entity_versions.html
```diff
@@ -21,14 +21,12 @@ <h2>Versions</h2>
 
   <div id="entityversionlist_content">
     <ul>
-    {% for tag, commit in versions %}
+    {% for tags, commit in versions %}
     <li>
       <p>
         <strong>
           <a class="entityversionlink" href="{{ entity|url_version:commit }}">
```
helenst: This part seems a little broken when I try it... I think the `url_version` filter is having problems. Needs a `str` around the `tags.get` call but also (if the intention is to link to each tag by its name - which seems like a nice idea) the url pattern will need adapting

weblab/entities/views.py
```diff
@@ -197,6 +198,48 @@ def get_redirect_url(self, *args, **kwargs):
...
+
+    def post(self, request, *args, **kwargs):
+        """Check the form and possibly add the tag in the repo.
+
+        Called by Django when a form is submitted.
+        """
+        import git
+        form = self.get_form()
+        entity = self.object = self.get_object()
+        if form.is_valid():
```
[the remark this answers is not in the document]
jonc125: I wondered about that. But I wasn't sure how to get at the specific git repo from there. Also, I don't think you want to add the tag within the clean function, which means there'd be a slight race condition between cleaning & adding where a tag with the same name could potentially be added. So you need to be able to report errors from the view anyway.

weblab/templates/entities/entity_versions.html
jonc125: I clearly wasn't thorough enough in my testing of nasty tag names 😞 I've now read `git check-ref-format --help` and tried to make this more robust. Hopefully I've covered every possibility!

jonc125:
Thanks for the feedback. Ready for re-review.

Issue #36, opened 2018-01-13
Only last of multiple uploaded files is kept


helenst:
If you upload multiple files when creating a new version of an entity, only the last one is kept!

As mentioned on #34

helenst:
This was fixed in #44

PR #44, opened 2018-01-25
Delete entities and files

### LINKED ISSUE #22: Allow deletion of entities and files
Deleting an entity should destroy the local repository (probably with a warning)
Deleting a file should be part of a new version of the model, with a corresponding commit.

helenst:
Start by clicking 'delete' icon on the version of an entity - it'll lead through to the 'new version' page. From there, other files can be deleted/added.

Fixes #22 
Also should take care of #36

### DIFF
--- a/weblab/entities/tests/test_views.py
+++ b/weblab/entities/tests/test_views.py
@@ -123,6 +123,8 @@ class TestEntityDeletion:
         entity = recipe.make(author=user)
         repo_path = entity.repo_abs_path
         add_version(entity)
+        assert Entity.objects.filter(pk=entity.pk).exists()
+        assert repo_path.exists()
 
         response = client.post(url % entity.pk)
 
@@ -237,17 +239,17 @@ class TestVersionCreation:
 
     def test_add_multiple_files(self, user, client):
         add_permission(user, 'create_model_version')
-        file1 = recipes.model_file.make(
-            entity__author=user,
+        model = recipes.model.make(author=user)
+        recipes.model_file.make(
+            entity=model,
             upload=SimpleUploadedFile('file1.txt', b'file 1'),
             original_name='file1.txt',
         )
-        file2 = recipes.model_file.make(
-            entity=file1.entity,
+        recipes.model_file.make(
+            entity=model,
             upload=SimpleUploadedFile('file2.txt', b'file 2'),
             original_name='file2.txt',
         )
-        model = file1.entity
 
         response = client.post(
             '/entities/models/%d/versions/new' % model.pk,
@@ -269,6 +271,7 @@ class TestVersionCreation:
         model = recipes.model.make(author=user)
         add_version(model, 'file1.txt')
         add_version(model, 'file2.txt')
+        assert len(model.repo.head.commit.tree.blobs) == 2
 
         response = client.post(
             '/entities/models/%d/versions/new' % model.pk,
@@ -283,6 +286,30 @@ class TestVersionCreation:
         assert 'delete-file' in model.repo.tags
         assert model.repo.head.commit.message == 'delete file1'
         assert len(model.repo.head.commit.tree.blobs) == 1
+        assert model.repo.head.commit.tree.blobs[0].name == 'file2.txt'
+
+    def test_delete_multiple_files(self, user, client):
+        add_permission(user, 'create_model_version')
+        model = recipes.model.make(author=user)
+        add_version(model, 'file1.txt')
+        add_version(model, 'file2.txt')
+        add_version(model, 'file3.txt')
+        assert len(model.repo.head.commit.tree.blobs) == 3
+
+        response = client.post(
+            '/entities/models/%d/versions/new' % model.pk,
+            data={
+                'delete_filename[]': ['file1.txt', 'file2.txt'],
+                'commit_message': 'delete files',
+                'version': 'delete-files',
+            },
+        )
+        assert response.status_code == 302
+        assert response.url == '/entities/models/%d' % model.id
+        assert 'delete-files' in model.repo.tags
+        assert model.repo.head.commit.message == 'delete files'
+        assert len(model.repo.head.commit.tree.blobs) == 1
+        assert model.repo.head.commit.tree.blobs[0].name == 'file3.txt'
 
     def test_delete_nonexistent_file(self, user, client):
         add_permission(user, 'create_model_version')
@@ -293,7 +320,7 @@ class TestVersionCreation:
             '/entities/models/%d/versions/new' % model.pk,
             data={
                 'delete_filename[]': ['file2.txt'],
-                'commit_message': 'delete file1',
+                'commit_message': 'delete file2',
                 'version': 'delete-file',
             },
         )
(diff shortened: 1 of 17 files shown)


jonc125:
I'm assuming this is ready to review @helenst ?

helenst:
@jonc125 Yes please :)

weblab/static/js/upload.js
```diff
@@ -58,6 +56,19 @@ Upload.prototype = {
...
+      if ($tr.hasClass('deleting')) {
+        $tr.removeClass('deleting');
+        $tr.find('input[name="delete_filename[]"]').remove()
+      } else {
+        $tr.addClass('deleting');
+        $('<input type="hidden" name="delete_filename[]" value="' + $.trim($td.text()) + '" />').appendTo($td);
+      }
+      return false;
+    });
   },
```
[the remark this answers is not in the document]
jonc125: Ah, that's OK because you don't show them the delete link for that file, just a 'cancel add' button effectively.

Interestingly, you can mark an existing file for deletion then upload a new file with the same name all within a single version change. Should we have a test that this does the right thing?

This is also possibly an unintended consequence of the fact that the check for a file upload having an existing name ignores files kept from previous versions - we should definitely fix that!

jonc125:
I added a couple of cosmetic changes to the branch as well.

weblab/entities/tests/test_views.py
```diff
@@ -203,6 +235,73 @@ def test_create_model_version(self, user, client):
...
+            data={
+                'delete_filename[]': ['file2.txt'],
+                'commit_message': 'delete file1',
+                'version': 'delete-file',
+            },
+        )
+        assert response.status_code == 302
+        assert response.url == '/entities/models/%d' % model.id
+        assert 'delete-file' in model.repo.tags
+        assert model.repo.head.commit.message == 'delete file1'
```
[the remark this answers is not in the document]
helenst: Agreed. It makes little sense to have a commit... I'll add a check.

weblab/entities/views.py
```diff
@@ -195,6 +196,21 @@ def get_redirect_url(self, *args, **kwargs):
         return reverse(url_name, args=[kwargs['pk'], 'latest'])
 
 
+class EntityDeleteView(UserPassesTestMixin, DeleteView):
+    """
+    Delete an entity
+    """
+    model = Entity
+    raise_exception = True
```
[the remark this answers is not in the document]
helenst: It's part of `UserPassesTestMixin` which causes the view to raise a 403 error if the user doesn't pass the test rather than redirecting to login. Which seems to make sense, since deletion links aren't provided for those users anyway, it's not something that's likely to happen as part of the usual flow - a 'forbidden' response seems more appropriate (and they may already be logged in anyway). I'll add a comment.

weblab/static/js/upload.js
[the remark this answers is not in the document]
helenst: I found a few bugs around marking for deletion and detecting existing uploaded files. It should now be possible to delete and replace a file in the same commit. (Might be useful if you have a local new copy of the file to upload - git will of course detect a deletion and re-addition as a change in the file)

helenst:
I think this has handled everything that was brought up, except for users acting simultaneously, which I've made into a separate issue for follow up work.

helenst:
i.e. ready for re-review @jonc125

weblab/entities/views.py
```diff
@@ -239,33 +242,63 @@ def get_context_data(self, **kwargs):
         kwargs['delete_file'] = self.request.GET.get('deletefile')
         return super().get_context_data(**kwargs)
 
-    def post(self, request, *args, **kwargs):
-        entity = self.get_object()
+    def add_and_delete_files(self, deletions, additions):
+        pass
```
jonc125: Was this supposed to be implemented? 😄

helenst: Oops... I had a refactor in mind but decided against it for now... Removed.

weblab/entities/views.py
```diff
@@ -239,33 +242,63 @@ def get_context_data(self, **kwargs):
...
-                           request.user.email)
-        entity.tag_repo(request.POST['version'])
-
-        return HttpResponseRedirect(
-            reverse('entities:%s' % entity.entity_type, args=[entity.id]))
+            except GitCommandError as e:
+                git_errors.append(e.stderr)
+
+        if git_errors:
+            # If there were any errors with adding or deleting files,
```
jonc125: Will Django do something sensible with the temporary upload files in this case, keeping them for a resubmission attempt?

Also, what happens if the user abandons the new version? Does Django clean up automatically after an interval?

weblab/static/js/upload.js
```diff
@@ -59,17 +56,30 @@ Upload.prototype = {
 
     $("form #entityversionfilestable").on('click', 'tr .action .delete-file', function() {
       var $tr = $(this).parents('tr');
+      return false;
+    });
+  },
+
+  toggleDelete: function($row) {
```
jonc125: Should this be `$tr`?

weblab/static/js/upload.js
```diff
@@ -59,17 +56,30 @@ Upload.prototype = {
 
     $("form #entityversionfilestable").on('click', 'tr .action .delete-file', function() {
       var $tr = $(this).parents('tr');
+      return false;
```
jonc125: This function doesn't appear to be doing anything?

jonc125: If I make it call `toggleDelete($tr)` (fixing the error below) then I can't delete multiple files in one go, because clicking the delete icon performs a GET request. (Why it should do this when the handler returns false I'm not sure!) If I leave it as-is, nothing happens at all when I click the delete icon.

helenst:
Argh - sorry. I did a little refactoring and thought it was all still working but it hadn't actually rebuilt the js. Sorry about that.

I've fixed this now. Turns out js errors were happening on click, so the handler couldn't get as far as returning false. I've fixed those. Can you see if it's all OK for you now?

weblab/entities/views.py
jonc125: @helenst can you answer the above queries?

jonc125:
Well, deleting multiple files now works.

Alas, if I (accidentally in this case!) use an existing tag as the version field, I get a git command error traceback from the tag command. But by this point the commit has already happened. So I think we need a try-catch for this that moves HEAD back before the commit...

weblab/entities/views.py
helenst: The temporary upload files will stay on the system (they're associated with the entity upon upload). There isn't a clean up yet, but probably should be. I'm not sure exactly at what point to decide it's appropriate to clean up an upload, since abandoning a version isn't a positive action by the user. Nightly job that cleans up anything older than 24 hours perhaps?

helenst:
@jonc125 That's a sensible idea. Will do...

weblab/entities/views.py
jonc125: That sounds sensible.

jonc125:
Works nicely. Now good to merge.

PR #49, opened 2018-04-06
Add spec docs for the experiment runner service

jonc125:
Hopefully these contain everything you need @helenst. Let me know if anything is unclear or missing!


jonc125:
@MichaelClerx has asked if running experiments can report on partial progress, by sending back the stdout thus far periodically with a 'running' status update message. Would this be easy to add in @helenst ? We could get them to package it up in a COMBINE Archive, so they same code can be reused, and subsequent messages just replace the previous set of files?

mirams:
yeah, would be nice

helenst:
This should be relatively straightforward, it'll need some extra logic but not a huge amount.

jonc125:
Great, I'll update the spec docs accordingly!

PR #51, opened 2018-04-25
Support experiments [WIP]

### LINKED ISSUE #15: Implement experiment matrix view
With the ability to trigger new experiment runs just from this interface at present.

Child of #2.

jonc125:
I'm creating a PR so I have somewhere to stick comments as I start looking at this :)

Fixes #15.

Things to do noticed while playing with this deployed on scrambler:
- [x] An 'Experiments' link is needed on the top menu
- [x] When the system fails to submit an experiment, it should set the status to 'failed' not 'inapplicable'
- [x] Once an experiment is submitted, clicking on the matrix square should take you to the entity page for that experiment without having to refresh the matrix
- [x] Being able to see the list of versions of an experiment
- [x] Being able to rerun an experiment (useful when it failed because the deployment isn't working yet!)
- [x] The experiments view should be accessible without login, but should only show those entities & experiments you have permission to see
- [ ] Running experiments with non-public entities is broken - when the runner tries to get the entity archive, it's redirected to the login page...
    - @helenst Could we do something like pass to the runner service (e.g. as part of the signature) a token which it could pass as part of get requests to log in automatically as a special user with read-only permission on every entity? Would this approach also secure the callback handler sufficiently?
- [x] All URLs seem to be of the form https://scrambler.cs.ox.ac.uk/experiments/10/versions/10, with version id matching experiment id, and that id showing as the version when you view the results. Is this intentional? Seems like starting the version id at 1 for each experiment would be better?
    - Ah, it's because these IDs are table row IDs, unrelated to git, of course. They won't always match, once we start having repeat versions etc. But we probably should improve the display
- [x] Related to this, I've just realised `Experiment` links to `model` and `protocol` while `ExperimentVersion` links to versions of the above. In WL1 `Experiment` links to `model_version` and `protocol_version`, and `ExperimentVersion` lets you run repeats of this configuration. I wonder which is better...
    - WL1: clear distinction between repeats of the same experiment, and what is actually a different experiment because it uses a different protocol/model (even if it's just another version, it's still different)
    - Current WL2: shows there's a relationship between experiments run with related entity versions. But perhaps hard to decide what the 'latest' version is?


helenst:
Thanks @jonc125 I'm just working on the view of the experiment files at the moment, hooking things up to the javascript visualisations etc. When that's done I'll need to do some tidying up and this will help - thanks!

weblab/entities/tests/test_models.py
```diff
@@ -1,48 +1,50 @@
...
-    other_user = recipes.user.make()
     superuser = recipes.user.make(is_superuser=True)
 
     assert model.is_deletable_by(user)
     assert model.is_deletable_by(superuser)
     assert not model.is_deletable_by(other_user)
+
+
+@pytest.mark.django_db
+def test_visibility():
```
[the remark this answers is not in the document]
helenst: It's testing at a different level (model vs view) although I think this one does now seem unnecessary.

weblab/entities/views.py
```diff
@@ -400,6 +404,33 @@ class ProtocolEntityVersionListView(ProtocolEntityTypeMixin, VersionListView):
...
+
+        zipfile_name = os.path.join(
+            get_valid_filename('%s_%s.zip' % (entity.name, commit.hexsha))
+        )
+
+        archive = entity.repo.archive(ref)
+
+        response = HttpResponse(content_type='application/zip')
+        response['Content-Disposition'] = 'attachment; filename=%s' % zipfile_name
+        response.write(archive.read())
```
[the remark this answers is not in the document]
helenst: I imagine it may cause a problem if datasets get large. Though we're only using this method for model and protocol downloads, as experiments are stored differently (i.e. already as a zip, so can be streamed directly from disk).

weblab/config/settings/base.py
```diff
@@ -53,6 +53,7 @@
 LOCAL_APPS = [
     'accounts.apps.AccountsConfig',
     'entities.apps.EntitiesConfig',
+    'experiments',
```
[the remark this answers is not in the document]
helenst: Because it doesn't have a custom app config set up... I learned a neater way to set these up though (in the app's `__init__.py`) so I'll change it to avoid confusion.

weblab/entities/views.py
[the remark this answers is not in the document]
jonc125: OK, that'll be fine then! Models & protocols shouldn't have large files in them.

weblab/experiments/forms.py
```diff
@@ -0,0 +1,11 @@
...
+
+from .models import ExperimentVersion
+
+
+class ExperimentSimulateCallbackForm(forms.ModelForm):
+    upload = forms.FileField(required=False)
+
+    class Meta:
+        model = ExperimentVersion
+        fields = ('status', 'return_text', 'task_id', 'upload')
```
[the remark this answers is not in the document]
helenst: This is for development purposes / possibly also for manual intervention by an admin user. It simulates the experiment result callback. I'm going through and adding more comments in various places, and will include this.

weblab/experiments/models.py
```diff
@@ -0,0 +1,110 @@
...
+
+from django.conf import settings
+from django.db import models
+
+from core import visibility
+from entities.models import ModelEntity, ProtocolEntity
+
+
+class Experiment(models.Model):
+    creation_date = models.DateTimeField(auto_now_add=True)
```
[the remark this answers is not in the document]
helenst: It seems like a good idea. I'm trying to think of a good name for this :smile:

weblab/experiments/processing.py
```diff
@@ -0,0 +1,151 @@
...
+    experiment, _ = Experiment.objects.get_or_create(
+        model=model,
+        protocol=protocol,
+        defaults={
+            'author': user,
+            'visibility': visibility.get_joint_visibility(model.visibility, protocol.visibility)
+        }
+    )
+
+    version = ExperimentVersion.objects.create(
```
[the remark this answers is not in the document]
helenst: I've added this

weblab/experiments/models.py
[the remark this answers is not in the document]
jonc125: `EntityBase` ?

jonc125:
One todo noticed while playing with this deployed on scrambler:
- [ ] An 'Experiments' link is needed on the top menu

weblab/conftest.py
```diff
@@ -20,9 +23,30 @@ def add_version(entity, filename='file1.txt', tag_name=None):
...
 
 
+@pytest.fixture(autouse=True)
+def fake_upload_path(settings, tmpdir):
+    settings.MEDIA_ROOT = str(tmpdir)
+    return settings.MEDIA_ROOT
+
+
+@pytest.fixture(autouse=True)
+def fake_experiment_path(settings, tmpdir):
```
jonc125: Something I just thought of - do these separate uses of `tmpdir` get different temporary folders?

MichaelClerx: If that becomes an issue, we can use the `tempdir` module:

- https://docs.python.org/2/library/tempfile.html
- https://github.com/pints-team/pints/blob/master/test/shared.py#L135-L174

weblab/.coveragerc
```diff
@@ -0,0 +1,10 @@
+[run]
```
jonc125: What service do we need to enable to see the results of this in GitHub?

weblab/conftest.py
jonc125: https://docs.pytest.org/en/latest/tmpdir.html isn't entirely clear, but I think they're separate!

weblab/.coveragerc
MichaelClerx: http://codecov.io/ ?

jonc125: I've started this off, but it needs some config in the `.travis.yml` too; will get to that next week probably!

weblab/conftest.py
MichaelClerx: Sorry, I meant the `tempfile` module! The other code I posted is just a ContextManager I wrapped around it

weblab/experiments/tests/test_views.py
```diff
@@ -1,81 +1,117 @@
...
 
 
 @patch('requests.post', side_effect=mock_submit)
 @pytest.mark.django_db
 class TestNewExperimentView:
     @pytest.mark.usefixtures('logged_in_user')
-    def test_submits_experiment(self, mock_post, client, model_with_version, protocol_with_version):
+    def test_submits_experiment(
+        self, mock_post,
+        client, logged_in_user, model_with_version, protocol_with_version
```
jonc125: Presumably we now don't need the `usefixtures` call before the method?

weblab/experiments/tests/test_views.py
```diff
@@ -1,81 +1,117 @@
...
+        assert data['newExperiment']['response']
 
         version = ExperimentVersion.objects.get()
 
         assert data['newExperiment']['expId'] == version.experiment.id
         assert data['newExperiment']['versionId'] == version.id
         assert data['newExperiment']['expName'] == version.experiment.name
-        assert data['newExperiment']['response']
 
+    @pytest.mark.usefixtures('logged_in_user')
```
jonc125: Similarly here?

weblab/experiments/views.py
```diff
@@ -146,16 +175,112 @@ def get_success_url(self):
...
-
         data = dict(
-            signature=self.kwargs['pk'],
-            **form.data
+            signature=self.get_object().signature,
+            **form.cleaned_data
         )
 
         result = process_callback(data, {'experiment': form.files.get('upload')})
+        print(result)
```
jonc125: Make sure to remove this when you're done debugging :)

weblab/static/js/experiment.js
```diff
@@ -0,0 +1,995 @@
+var showdown = require('showdown');
+var notifications = require('./lib/notifications.js');
+var utils = require('./lib/utils.js')
+var expt_common = require('./expt_common.js');
+var plugins = [
+  //require('./visualizers/displayBivesDiff/displayBivesDiff.js'),
+  require('./visualizers/displayContent/displayContent.js'),
+  require('./visualizers/displayPlotD3/displayPlotD3.js'),
```
jonc125: I should have said this before actually - we're deprecating this visualiser, so if anything in it breaks just delete it!

weblab/static/js/main.js
```diff
@@ -1,80 +1,20 @@
 var $ = require('jquery');
+require('jquery-migrate');
+$.migrateMute = true;
 require('jquery-ui-browserify');
-var XDate = require('xdate');
 var utils = require('./lib/utils.js')
```
jonc125: Is there a missing ; here?

weblab/static/js/main.js
```diff
@@ -269,6 +150,10 @@ function initPage ()
       }
     }
   });
+
+  if ($('#experiment-version').length > 0) {
```
jonc125: This feels like quite a hack - is there a better way provided by `require` for initialising other modules?

weblab/core/visibility.py
```diff
@@ -37,6 +38,17 @@ def get_joint_visibility(*visibilities):
     )]
 
 
+def visibility_query(user):
```
helenst: Great, this is much nicer here than on `User`

weblab/experiments/models.py
```diff
@@ -66,12 +101,18 @@ class ExperimentVersion(UserCreatedModelMixin, VisibilityModelMixin, models.Mode
...
     task_id = models.CharField(max_length=50, blank=True)
-    model_version = models.CharField(max_length=50)
-    protocol_version = models.CharField(max_length=50)
 
     def __str__(self):
         return '%s at %s: (%s)' % (self.experiment, self.created_at, self.status)
 
+    @property
+    def name(self):
+        return str(self.experiment.versions.filter(pk__lte=self.pk).count())
```
helenst: Comparing the `pk` worries me a little - can it always be relied on that older pks are lower? Using `created_at` might be more robust.

jonc125: I was assuming that `pk` would be an auto-increment field so this would be safe. But using `created_at` works too!

helenst:
@jonc125 I'm happy for this to be merged whenever you are :+1:

jonc125:
Great! Created #55 for the last item above.

Issue #54, opened 2018-05-24
Experiment result comparison views


jonc125:
There are several parts to this, in priority order:
- [x] The view to compare two or more experiments, e.g. like https://travis.cs.ox.ac.uk/FunctionalCuration/compare/e/9081/5640/5491
- [x] The 'comparison mode' functionality for the matrix that lets you get to these views easily
- [ ] Getting to a comparison view from a model or protocol version view, via the 'Compare protocols' / 'Compare models' buttons (e.g. https://travis.cs.ox.ac.uk/FunctionalCuration/model/BeelerReuter1977/4/v3otha/151 ) - see #58.

Probably best to tackle each part in a separate PR.

Child of #2.

PR #57, opened 2018-06-08
Experiment comparison

helenst:
Experiment comparison view (first point of #54). Not linked from anywhere yet.

Example URL:
`http://weblab.localhost:8000/experiments/compare/80/81/82/`

where 80, 81, 82 are experiment version IDs to be compared.

### DIFF
--- a/weblab/experiments/views.py
+++ b/weblab/experiments/views.py
@@ -177,9 +177,9 @@ class ExperimentComparisonView(TemplateView):
     template_name = 'experiments/experimentversion_compare.html'
 
     def get_context_data(self, **kwargs):
-        versions = ExperimentVersion.objects.visible_to(self.request.user).filter(
-            pk__in=map(int, self.kwargs['version_pks'][1:].split('/')),
-        ).order_by('created_at')
+        pks = {int(pk) for pk in self.kwargs['version_pks'][1:].split('/')}
+        versions = ExperimentVersion.objects.visible_to(
+            self.request.user).filter(pk__in=pks).order_by('created_at')
 
         kwargs.update({
             'experiment_versions': versions,
@@ -246,17 +246,27 @@ class ExperimentComparisonJsonView(View):
         }
 
     def get(self, request, *args, **kwargs):
-        versions = ExperimentVersion.objects.visible_to(self.request.user).filter(
-            pk__in=map(int, self.kwargs['version_pks'][1:].split('/')),
-        ).order_by('created_at')
+        pks = {int(pk) for pk in self.kwargs['version_pks'][1:].split('/')}
+        versions = ExperimentVersion.objects.visible_to(
+            self.request.user).filter(pk__in=pks).order_by('created_at')
 
-        return JsonResponse({
+        response = {
             'getEntityInfos': {
                 'entities': [
                     self._version_json(version) for version in versions
                 ]
             }
-        })
+        }
+
+        if len(versions) <= len(pks):
+            response['notifications'] = {
+                'errors': [
+                    'Some requested experiment results could not be found '
+                    '(or you don\'t have permission to see them)'
+                ]
+            }
+
+        return JsonResponse(response)
 
 
 @method_decorator(staff_member_required, name='dispatch')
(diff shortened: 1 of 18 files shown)


jonc125:
Funky, I didn't realise codecov would fail checks for low coverage :)

MichaelClerx:
Yup! It fails PRs that would decrease overal coverage!

(So you can cheat by adding tests for other things :D )

helenst:
OK - it's got enough tests to make codecov happy - ready to review! :smile:

weblab/experiments/views.py
```diff
@@ -170,6 +170,93 @@ class ExperimentVersionListView(VisibilityMixin, DetailView):
...
 
+class ExperimentComparisonView(TemplateView):
+    """
+    Compare multiple experiment versions
+    """
+    template_name = 'experiments/experimentversion_compare.html'
+
+    def get_context_data(self, **kwargs):
+        versions = ExperimentVersion.objects.filter(
+            pk__in=map(int, self.kwargs['version_pks'][1:].split('/'))
```
jonc125: Should we give some sort of error if the user requests non-existent versions, or versions they're not allowed to view? Or just silently filter so it only shows ones they can see?

I think my preferred option would be to compare the versions they can see (if any) but add an error notification "Some requested experiment results could not be found (or you don't have permission to see them)"

weblab/experiments/views.py
```diff
@@ -170,6 +170,93 @@ class ExperimentVersionListView(VisibilityMixin, DetailView):
...
+            'download_url': reverse(
+                'experiments:archive', args=[version.experiment.id, version.id]
+            ),
+            'modelName': version.experiment.model.name,
+            'protoName': version.experiment.protocol.name,
+        }
+
+    def get(self, request, *args, **kwargs):
+        versions = ExperimentVersion.objects.filter(
+            pk__in=map(int, self.kwargs['version_pks'][1:].split('/'))
```
jonc125: Similarly to above, we should filter this list so only visible versions are returned.

weblab/static/js/compare.js
```diff
@@ -284,7 +262,8 @@ function parseEntities (entityObj)
 				var file = entity.files[j],
 					sig = file.name.hashCode();
 				file.signature = sig;
-				file.url = contextPath + "/download/" + entityType.charAt(0) + "/" + convertForURL (entity.name) + "/" + entity.id + "/" + file.id + "/" + convertForURL (file.name);
+				file.url = file.url;
```
jonc125: Probably don't need this line? :)

weblab/static/js/compare.js
```diff
@@ -510,8 +492,19 @@ function handleReq ()
...
+
+  $.getJSON(url, function(data) {
+    notifications.display(data);
+    gotInfos = true;
+
+    if (data.getEntityInfos) {
+      parseEntities(data.getEntityInfos.entities);
+    }
+  })
+  /*
```
jonc125: Why not just delete the old-style call?

helenst:
@jonc125 You're right - I've restricted to visible experiments, and addressed your other comments. Ready for re-review.

weblab/static/js/compare.js
jonc125: Looks good, apart from this fix is missing :)

helenst: d'oh! Fixed.

weblab/experiments/views.py
```diff
@@ -258,7 +258,7 @@ def get(self, request, *args, **kwargs):
             }
         }
 
-        if len(versions) <= len(pks):
+        if len(versions) < len(pks):
```
jonc125: Good spot. However, testing it here I find it now never shows! I think this is because this check is in the JSON view, which only gets the visible version IDs in its URL, since the main view has already filtered them. So I think we need the main view to fill in the `notes.errors` context variable, and the `weblab/templates/includes/notifications.html` template should display these (at present that loop is commented out).

helenst: Good point - that never got converted over from the original JSP. I think it's doing the right thing now!

jonc125:
Good to merge now.

PR #59, opened 2018-06-14
Comparison matrix

helenst:
Enable and fix comparison interface on experiment matrix, along with click through to submatrix views

Part of #54.

### DIFF
--- a/weblab/experiments/tests/test_views.py
+++ b/weblab/experiments/tests/test_views.py
@@ -57,6 +57,19 @@ class TestExperimentMatrix:
         assert str(exp.protocol.pk) in data['getMatrix']['protocols']
         assert str(exp.pk) in data['getMatrix']['experiments']
 
+    @pytest.mark.usefixtures('logged_in_user')
+    def test_experiment_json(self, client, experiment_version):
+        exp = experiment_version.experiment
+
+        response = client.get('/experiments/matrix')
+        data = json.loads(response.content.decode())
+
+        exp_data = data['getMatrix']['experiments'][str(exp.pk)]
+        assert exp_data['id'] == experiment_version.id
+        assert exp_data['entity_id'] == exp.id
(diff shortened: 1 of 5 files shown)


weblab/experiments/tests/test_views.py
```diff
@@ -75,6 +106,51 @@ def test_anonymous_cannot_see_private_data(self, client, experiment_version):
...
+                'protoIds[]': [experiment_version.experiment.protocol.pk, non_existent_pk],
+            }
+        )
+
+        assert response.status_code == 200
+        data = json.loads(response.content.decode())
+        assert 'getMatrix' in data
+        assert len(data['getMatrix']['models']) == 1
+        assert len(data['getMatrix']['protocols']) == 1
+        assert len(data['getMatrix']['experiments']) == 1
```
jonc125: We should probably check that we get back the ids we're expecting here.

weblab/experiments/views.py
```diff
@@ -93,9 +106,13 @@ def get(self, request, *args, **kwargs):
         # Only give info on experiments involving the correct entity versions
         experiments = {}
         for exp in Experiment.objects.filter(model__in=q_models, protocol__in=q_protocols):
-            if (exp.model_version == models[exp.model.pk]['version'] and
-                    exp.protocol_version == protocols[exp.protocol.pk]['version']):
-                experiments[exp.pk] = self.experiment_json(exp)
+            if ((exp.model_version == models[exp.model.pk]['version'] and
```
jonc125: Why the double (())?

helenst: pep8-friendly indentation fix to prevent confusing alignment in multiline if ... although I'm not actually sure why I changed this, since the original code handles this, just in a different way (extra indent level for second part of clause) :smile:

jonc125:
Great!

PR #60, opened 2018-06-15
Use page-specific titles

jonc125:
To make the browser back button menu more informative!

### DIFF
--- a/weblab/templates/entities/entity_list.html
+++ b/weblab/templates/entities/entity_list.html
@@ -1,6 +1,8 @@
 {% extends "base.html" %}
 {% load entities %}
 
+{% block title %}My {{ type }}s - {% endblock title %}
+
 {% block content %}
 
   <a class="button {% if type == 'model' %}selected{% endif %}" href="{% url "entities:models" %}" id="modelchooser">models</a>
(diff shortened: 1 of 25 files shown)


helenst:
Coverage shouldn't be complaining since this only affects template files. Might be because it needs rebasing... trying that now.

jonc125:
Nice! I don't quite understand why codecov complains some times - it seems even adding new fully covered code can sometimes decrease overall coverage?

weblab/templates/entities/entity_list.html
```diff
@@ -1,6 +1,8 @@
 {% extends "base.html" %}
 {% load entities %}
 
+{% block title %}My {{ type }}s - {% endblock title %}
```
helenst: Title uses "My" but page heading uses "Your". We should pick one :) The interface isn't very consistent on this front anyway (has both my files / your files) - carried over from v1 I think so might need some thought as to which we use.

jonc125: Good point! I think 'your' is probably better. Do @mirams @MichaelClerx have an opinion?

MichaelClerx: I prefer 'your' !
Alternatively: <user name>'s

PR #72, opened 2018-08-08
Stop notifications from moving page context

### LINKED ISSUE #70: Don't make notes/error messages push page content down
This is particularly annoying for the matrix page: if you try and click across a row to start some experiments it shifts the matrix down and messes up the mouse clicking!

Options are:
1. Put the box at the bottom of the page content. Possibly not ideal as people might not notice it!
1. Make the box sit on top of the page content, so it doesn't move it, but still at the top. You could still get issues if the box gets so big it covers the part you're clicking on, so maybe it should have a maximum size with a 'show all' button as well as the 'dismiss' link?

Have you a preference @mirams ?

jonc125:
Fixes #70. I've gone for the simple approach of just making the notification box sit on top of the main content. It will grow indefinitely as new errors/notes are added, until dismissed.

### DIFF
--- a/weblab/static/sass/style.scss
+++ b/weblab/static/sass/style.scss
@@ -7,6 +7,8 @@
 @import "displayTable/displayTable";
 @import "editMetadata/editMetadata";
 
+$page-width: 850px;
+
 body {
 	font-family: 'Open Sans', sans-serif;
 	font-size: 16px;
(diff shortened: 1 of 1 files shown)


jonc125:
I experimented with `opacity: 0.75;` etc to make it semi-transparent, and found it just looked distracting/confusing unless you went to 0.99, at which point you couldn't see what was underneath at all!

helenst:
This looks great! :+1:

Issue #73, opened 2018-08-13
Review speed of experiments matrix view


jonc125:
As we get more experiments on scrambler, we should check it's presenting the matrix at least as quickly as WL1 could.

Child of #68.

jonc125:
And indeed it's now falling over entirely due to `OSError: [Errno 24] Too many open files` - see https://rollbar.com/jonc125/Web-Lab/items/9/

jonc125:
Looks like there are known issues here: https://gitpython.readthedocs.io/en/stable/intro.html#leakage-of-system-resources

MichaelClerx:
That's disappointing! Wonder why they put up a message apologising for the bug instead of fixing it...

MichaelClerx:
Not entirely sure what they're talking about. This offers some hints: https://stackoverflow.com/questions/14628486/why-arent-destructors-guaranteed-to-be-called-on-interpreter-exit

mirams:
This is really slow - just timed (when logged in as myself, in case that matters):
on `scrambler` web lab it takes 18 seconds from pressing the 'Experiment' tab to seeing the matrix, on the old `travis` it takes less than 1 second.

mirams:
Dropped down to no use, and still takes about 4 seconds rather than less than 1 :-(

jonc125:
Currently we have nearly 7s for the public view:
```
May 09 12:04:46 scrambler weblab[11258]: [pid: 11365|app: 0|req: 9/14] 128.40.131.160 () {50 vars in 1096 bytes} [Thu May  9 11:04:39 2019] GET /experiments/matrix?subset=public => generated 655315 bytes in 6864 msecs (HTTP/2.0 200) 3 headers in 104 bytes (5 switches on core 0)
```

jonc125:
On my local deploy, with fewer experiments, 1637 msecs:
```
May 09 11:20:14 weblab weblab[357]: [pid: 382|app: 0|req: 5/6] 10.0.2.2 () {46 vars in 1050 bytes} [Thu May  9 11:20:12 2019] GET /experiments/matrix?subset=mine => generated 13451 bytes in 1637 msecs (HTTP/1.1 200) 4 headers in 117 bytes (1 switches on core 0)
```
So still slow enough to see improvements! To produce this it does 109 SELECT queries and 10 calls to the git executable...

jonc125:
[...]

Issue #68, opened 2018-08-02
Achieve WL1 parity


jonc125:
At least for features we still care about, so we can move all entities over and turn off the old WL1 (perhaps after a period where it's read-only - what do you reckon @MichaelClerx @mirams ?).

Features needed in approximate priority order:
- Parent of #28
- Parent of #41
- Parent of #40
- Parent of #26
- Parent of #65
- Parent of #66
- Parent of #63
- Parent of #64
- Parent of #42

Optional features (that we could perhaps live without initially?):
- Parent of #39
- Parent of #67
- Parent of #48

jonc125:
All issues linked to this have now been closed!

mirams:
Does that mean we should be migrating over to the new server and retiring travis' web lab (Fergus wants to upgrade the machine)?

jonc125:
Yes, we should plan for that. I need to merge another PR and deploy on scrambler, then double-check it's happy, but at that point we (you?) should probably reach out to known users and ask them to try it out.

Issue #41, opened 2018-01-22
Implement per-version visibility


jonc125:
In the old Web Lab visibility is a per-entity-version setting, rather than the per-entity setting we have at present. This is valuable as it lets people keep the 'under construction' version of a model/protocol private and just publish the latest version.

The default visibility for a new version is to be the same as the previous version (parent commit).

Exactly how to implement this is tricky! (Cf discussion in #23) Thoughts welcome @helenst :)

Slightly related to #40 (notably the 'moderated' visibility setting), but tangential.

jonc125:
[git notes](http://alblue.bandlem.com/2011/11/git-tip-of-week-git-notes.html) looks like a promising approach.

MichaelClerx:
If we're storing this data in the notes, what does that mean for someone who clones and modifies the repo? Will they be able to break our system (by accident) or need to know about internal conventions we make?

jonc125:
Notes aren't cloned by default, so someone would have to be quite deliberate about trying to confuse the system (and have push access (or the ability to ask the Web Lab to pull) so could alter the visibility through the Web Lab directly anyway). You can also put your notes in a non-default location so the normal `git notes show` command doesn't pick them up. So I think we're fairly safe!

jonc125:
Expanding a bit on the example above to store notes in our own location:
```python
prefix = 'Visibility: '
# Overwrite existing note to store new visibility
cmd.notes('--ref', 'weblab', 'add', '-f', '-m', prefix+'public', commit.hexsha)
# Read visibility for a commit
try:
    note = cmd.notes('--ref', 'weblab', 'show', commit.hexsha)
except git.exc.GitCommandError:
    note = prefix + 'private'
if not note.startswith(prefix):
    note = prefix + 'private'
visibility = note[len(prefix):]
```

helenst:
Git notes look neat, and worth knowing about!

One thing worth thinking about is that a note applies to a single commit, and doesn't automatically carry on through to the next one. But we need to be able to know the visibility of any commit/version.

Options here:

(a) add a visibility note to every version regardless of whether the visibility changed. Do-able in weblab, but easily forgotten on manual versions.

(b) add visibility note only when the visibility changes. We'd need to be able to backtrack through the parent commits of any given version to find out the last visibility note. In a large, old repo that could be quite a lot of backtracking and might impact performance (especially if we want to show this info on the version listing page, for example)

[...]

PR #76, opened 2018-08-13
Improve performance of matrix view

jonc125:
Work on #73. Basically, accessing the git repos is slow, so this tries to avoid doing so where possible. I've also avoided caching the repo wrapper class, which will hopefully avoid running out of file handles! It's certainly made an improvement on scrambler - not quite as fast as WL1 (which can generate the view purely from DB queries) but much closer. I think now it only hits the filesystem for the row & column labels, not for each experiment.

We may wish to think @helenst about caching some bits of git metadata within the DB for speed - perhaps in a later phase once features for #68 are done. I'm thinking the list of commit SHAs and tags, perhaps also visibility stuff for #41 ? Provided all updates to the local repos on the server are mediated through Django (i.e. we don't allow direct push) it should be possible to keep things in sync, and we could have a cron job that re-syncs just in case.

### DIFF
--- a/requirements/Makefile
+++ b/requirements/Makefile
@@ -1,12 +1,14 @@
 .PHONY: all check clean
 
+upgrade_flag = #-U
+
 objects = $(wildcard *.in)
 outputs := $(objects:.in=.txt)
 
 all: $(outputs)
 
 %.txt: %.in
-	pip-compile -v --output-file $@ $<
+	pip-compile -v $(upgrade_flag) --output-file $@ $<
 
 test.txt: base.txt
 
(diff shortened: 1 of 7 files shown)


MichaelClerx:
Caching sounds like a Good Idea! From the current WL it seems to me like look-up is much more common than adding / modifying, so would be good to speed this up. Also adding/modifying is probably already relatively slow, so that also updating a cache won't hurt its performance

helenst:
Great! These changes to the repo lookups give significant performance improvements on that page. I've added a little further optimization to reduce the query count from the json view, but latest commit lookups on entities seem to be the next bottleneck to tackle. Caching these seems like a good idea - I think direct push could still be possible if we have some kind of process watching the filesystem for changes, but I'm guessing not a high priority at this stage.

PR #79, opened 2018-08-15
Use npm FileSaver library to restore export-plot-data functionality

### LINKED ISSUE #78: The 'export plot data' button seems to be broken
From Chrome's developer tools:
```
Uncaught ReferenceError: saveAs is not defined
    at HTMLAnchorElement.<anonymous> (main.js:sourcemap:3868)
    at HTMLAnchorElement.dispatch (main.js:sourcemap:29369)
    at HTMLAnchorElement.elemData.handle (main.js:sourcemap:29177)
```
Called at line 307 of `expt_common.js`.

WL1 used some third party JS to enable this: res/js/3rd/FileSaver/Blob.js and res/js/3rd/FileSaver/FileSaver.js

Child of #68.

jonc125:
Fixes #78.


weblab/static/js/expt_common.js
```diff
@@ -284,6 +285,11 @@ function getKeyValues(file, numTraces)
  */
 function allowPlotExport(filename, datasets, axisLabels)
 {
+    try {
+        var isFileSaverSupported = !!new Blob;
+    } catch (e) {
+        return; // No saveAs() available
+    }
```
helenst: Is this in case of the browser not supporting the FileSaver library?

jonc125: Yes, that's the check the FileSaver docs recommend.

PR #83, opened 2018-10-30
[WIP] Per-version entity visibility

### LINKED ISSUE #41: Implement per-version visibility

### LINKED ISSUE #63: Allow entity version visibility to be changed after creation
In WL1 this is done via making the visibility display in the entity version view a drop-down menu (assuming you have write permission for the entity) from which you can select a new visibility. Doing so triggers an AJAX call to make the change.

Depends on #41.

### LINKED ISSUE #73: Review speed of experiments matrix view

### LINKED ISSUE #81: Improve git performance with caching
Child of #68, following on from #76. Comments from there:

Me: We may wish to think about caching some bits of git metadata within the DB for speed - perhaps in a later phase once features for #68 are done. I'm thinking the list of commit SHAs and tags, perhaps also visibility stuff for #41 ? Provided all updates to the local repos on the server are mediated through Django (i.e. we don't allow direct push) it should be possible to keep things in sync, and we could have a cron job that re-syncs just in case.

Helen: Great! These changes to the repo lookups give significant performance improvements on that page. I've added a little further optimization to reduce the query count from the json view, but latest commit lookups on entities seem to be the next bottleneck to tackle. Caching these seems like a good idea - I think direct push could still be possible if we have some kind of process watching the filesystem for changes, but I'm guessing not a high priority at this stage.

helenst:
Probably still a bit of tidying up to do (but comments welcome at this stage) - creating a PR now so I can check test coverage.

This fixes #41 but also fixes #63 and fixes #81 and fixes #73 (for now at least). Related to #59 

Stores visibility as a `git note` on the version's commit.

Information about the contents of an entity's repository (commit IDs, timestamps, tags and visibility data) is primarily retrieved from the database and managed by the `repocache` module which provides models mirroring entities, versions and tags. The information is updated when it changes in the webapp, and a management command is provided for updating the repocache database tables. This should be run initially when this change is deployed:

```
python manage.py populate_entity_cache
```

Although technically this shouldn't need continually updating (at least while the webapp is the only thing modifying repositories) it could be run by a cron task, and it might be a good idea to have a button on the front end so an admin / entity owner can refresh the info from the repository.


jonc125:
One quick comment: I'd defer the admin button / cron task to another issue rather than making it part of this PR.

weblab/repocache/entities.py
```diff
@@ -0,0 +1,39 @@
...
+    Get IDs of all publicly visible entities
+
+    :return: set of entity IDs
+    """
+    return set(
+        CachedEntityVersion.objects.annotate(
+            latest_ts=Max('entity__versions__timestamp')
+        ).filter(
+            timestamp=F('latest_ts'),
+            visibility='public',
```
jonc125: How will this behave with an entity where (say) the latest version is private but the previous version is public?

Ideally you should still return this entity, but when other code displays a version it then they only show the public version, not the latest version. However, we could leave implementing that for a second PR, since it's probably non-trivial and not that common a use case.

weblab/repocache/exceptions.py
```diff
@@ -0,0 +1,3 @@
+
+class RepoCacheMiss(Exception):
```
jonc125: A docstring would be nice, explaining in what circumstances this can arise.

weblab/repocache/models.py
```diff
@@ -0,0 +1,114 @@
...
+
+    @property
+    def latest_version(self):
+        return self.versions.latest()
+
+    def get_version(self, sha):
+        """
+        Get a version of the entity
+
+        :param sha: hex string of the commit SHA of the version
```
jonc125: or 'latest'

weblab/repocache/models.py
```diff
@@ -0,0 +1,114 @@
...
+class CachedEntityTag(models.Model):
+    """
+    Cache for a tag in an entity's repository
+    """
+    entity = models.ForeignKey(CachedEntity, related_name='tags')
+    tag = models.CharField(max_length=255)
+    version = models.ForeignKey(CachedEntityVersion, related_name='tags')
+
+    class Meta:
+        unique_together = ['entity', 'tag']
```
jonc125: Does this automatically handle the case where a tag is updated to point at a newer version?

weblab/repocache/populate.py
```diff
@@ -0,0 +1,55 @@
...
+
+        # If the commit has no visibility info in the repo. remember this
+        # and wait until we find visibility on an earlier commit (which
+        # will be encountered later since we are iterating backwards through
+        # commits)
+        if visibility:
+            # Apply this commit's visibility to all those later (previously
+            # encountered) commits which have no visibility info
+            for sha in commits_without_visibility:
+                entity.set_version_visibility(sha, visibility)
```
jonc125: Presumably after this loop we should clear `commits_without_visibility`?

weblab/entities/migrations/0009_remove_entity_visibility.py
```diff
@@ -0,0 +1,19 @@
...
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('entities', '0008_auto_20180510_0824'),
+    ]
+
+    operations = [
```
jonc125: This looks like a likely culprit - no reverse migration!

weblab/repocache/migrations/0001_initial.py
```diff
@@ -0,0 +1,60 @@
...
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [
+        ('entities', '0009_remove_entity_visibility'),
```
jonc125: This looks like it would just lose all existing visibility info, rather than migrating the entity visibilities to be version visibilities? Which is probably fine at this early stage of the project, since we don't have much real visibility data, but may also have contributed to my errors.

jonc125:
It does! I can now see the matrix when logged in (using this branch). When logged out you get an empty matrix, presumably because all visibilities have defaulted to private. When I set one protocol & model version to public, they appear.

weblab/entities/models.py
```diff
@@ -64,6 +69,126 @@ def nice_version(self, commit):
...
+
+        :param commit: ref of the relevant commit
+        :param visibility: string representing visibility
+        """
+        commit = self.repo.get_commit(commit)
+        self.set_visibility_in_repo(commit, visibility)
+
+        try:
+            self.repocache.get_version(commit.hexsha).set_visibility(visibility)
+        except RepoCacheMiss:
```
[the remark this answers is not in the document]
helenst: It was due to some slight problems with the order in which things were done in the 'add new version' view. I've corrected this now.

weblab/entities/models.py
```diff
@@ -64,6 +69,126 @@ def nice_version(self, commit):
...
+    def get_version_visibility(self, sha):
+        """
+        Get the visibility of the given entity version
+
+        This is fetched from the repocache
+
+        :param sha: SHA of the relevant commit
+        :return: string representing visibility
+        """
+        return self.repocache.get_version(sha).visibility
```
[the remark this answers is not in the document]
helenst: I've not put miss handling directly in here because I wanted the views to have freedom in how they handle this. (e.g. a 404 is appropriate in some cases). I found some other places where it wasn't handled, so I've set up a `default` arg to deal with this.

weblab/repocache/entities.py
helenst: Yeah at the moment it won't include that entity at all (there's a test case for this). I'll leave it for now but that sounds like a sensible feature to add at some point.

weblab/repocache/models.py
helenst: I don't think we have this functionality in the webapp, but this should be easily done if it is added. (I'm not sure if I've understood this right)

weblab/entities/models.py
```diff
@@ -64,6 +69,126 @@ def nice_version(self, commit):
             version = version[:8] + '...'
         return version
 
+    def get_visibility_from_repo(self, commit):
+        """
+        Get the visibility of the given entity version from the repository
```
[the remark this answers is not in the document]
helenst: There are two place that use this:
1. `populate` command (which does do the backtracking through previous versions)
2. `add_version` when a new entity version is added. This calls `set_visibility_in_repo` first. So it should always be there. But I think setting `PRIVATE` would be a sensible default thing to do.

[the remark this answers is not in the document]
helenst: Ah... and I just remembered that `populate` is exactly why it returns `None` - it acts as an "unknown" visibility so we know to backtrack and add the visibility from a previous version.

weblab/entities/migrations/0009_remove_entity_visibility.py
helenst: It shouldn't have one as it's an auto generated schema migration - Django will know to add the field back in when reversing.

weblab/repocache/migrations/0001_initial.py
helenst: Yeah this is a tricky one as it involves moving data out of the database into somewhere external - it could be done, although I'm not sure it could be easily done in a migration - a script could be written if necessary though.

helenst:
> It does! I can now see the matrix when logged in (using this branch). When logged out you get an empty matrix, presumably because all visibilities have defaulted to private. When I set one protocol & model version to public, they appear.

Does this mean it's fully working now, or are you still having migration troubles?

jonc125:
> Does this mean it's fully working now, or are you still having migration troubles?

Migrating forwards seems to work, although I've not had time today to do any further experimentation. I haven't tried going back from here.

helenst:
OK @jonc125 I think I've addressed everything here, aside from any remaining migration problems

weblab/repocache/models.py
jonc125: Good point, we don't have the facility to delete tags (yet).

jonc125:
For potential future reference, I used the following to set all versions of all entities to public on scrambler, in `./manage.py shell --settings config.settings.deployed`:
```python
from entities.models import ModelEntity, ProtocolEntity
from core.visibility import Visibility

for m in ModelEntity.objects.all():
  for v in m.repo.commits:
    m.set_version_visibility(v.hexsha, Visibility.PUBLIC)

for m in ProtocolEntity.objects.all():
  for v in m.repo.commits:
    m.set_version_visibility(v.hexsha, Visibility.PUBLIC)
```

Issue #85, opened 2018-11-09
Broken links from entity versions list


jonc125:
E.g. in https://scrambler.cs.ox.ac.uk/entities/models/20/versions clicking any version link just changes the page URL without loading the new page.

This is because the entity.js Javascript is still being used for this page, and it defines `window.onpopstate` to call its `render()` function, which alters link handling.

That Javascript shouldn't be used for this page I think!

This is probably the same issue as #69.

Child of #68.

PR #86, opened 2018-11-16
Remove javascript for displaying entity version

### LINKED ISSUE #69: Bug viewing versions of experiments
In the experiment version list view, e.g. https://scrambler.cs.ox.ac.uk/experiments/18/versions/, clicking on the link for a version doesn't take you there - it changes the URL bar but doesn't visit the page.

There's a JS error:
```
ReferenceError: parseUrl is not defined
    at render (entity.js:832)
    at nextPage (entity.js:827)
```

Child of #68.

helenst:
Entity and experiment versions are accessed through full links, so remove the javascript that tries to handle a click event on those links. This should fix #69 #85.

### DIFF
--- a/weblab/static/js/entity.js
+++ b/weblab/static/js/entity.js
@@ -462,17 +462,6 @@ function registerFileDisplayer (elem)
     	}, true);
 }
 
-function registerVersionDisplayer (elem)
-{
-	elem.addEventListener("click", function (ev) {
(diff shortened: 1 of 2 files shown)


helenst:
I think `gulpfile.js` could be modified to build separate javascript bundles in addition to main.js, with each referencing the appropriate modules via `require`.

Issue #48, opened 2018-03-23
Implement protocol file syntax checking


jonc125:
This is also where the protocol interface gets stored within the DB. (Or in the future, in an RDF store.)

With additional features from the original:
* Create an error.txt file for any problems seen (the original does this too)
* Generate the readme.md as part of this step? (See also #75)
* Report in error.txt if there is no master file marked

We'll need to think about the most appropriate way of adding the error.txt file. Probably we have to create a new commit in the repository, and maybe move any tags? Or should it be a more ephemeral 'file' associated with the version only in our cache - since you wouldn't want it hanging around in a subsequent updated version!

jonc125:
@helenst I hope to have a bit of time to start on this next week. Do my ideas about making these ephemeral files in the repocache seem like a sensible first approach to try?

helenst:
We'd need to be careful about the cache repopulation script - currently it wipes everything from the cache and regenerates from scratch. Depending on whether that's actually being run (it shouldn't need to be). So if the repocache is the only place this file lives, it could be lost that way. 

One other possibility might be to store it as a git note?

jonc125:
Indeed, I'm thinking using a note with repocache for speed is probably the way to go. Will have a play in that direction :)

PR #98, opened 2018-12-19
Ephemeral files and protocol readmes

### LINKED ISSUE #75: Extract documentation from protocols as README.md
When a new protocol (version) is uploaded, the `documentation {}` block should be parsed out and used to create a README.md file within the repo, but only if a README.md file doesn't already exist.

This will also need to be done as a one-off batch migration operation for existing protocols.

#48 suggests doing this as part of the protocol file syntax checking - so the backend (which needs to parse the protocol anyway) can extract the block and return it. But really you don't want to create a new version for this, you want it as part of the initial upload version. So I think it needs to be done here. Or you need to create an ephemeral file-like object as suggested for the `errors.txt` file.

Child of #68.

jonc125:
This is work towards #75 and then #48. Will fix #75 when done.

### DIFF
--- a/weblab/conftest.py
+++ b/weblab/conftest.py
@@ -169,4 +169,4 @@ def logged_in_admin(client, admin_user):
 @pytest.fixture
 def model_creator(user, helpers):
     helpers.add_permission(user, 'create_model')
-    return User.objects.get(pk=user.pk)
+    return user
--- a/weblab/core/models.py
+++ b/weblab/core/models.py
@@ -59,6 +59,11 @@ class UserCreatedModelMixin(models.Model):
 
     @property
     def collaborators(self):
+        """
+        All users who have specifically been permitted to collaborate this object
+
+        :return list of `User` objects
+        """
         return [
             user
             for (user, perms) in get_users_with_perms(self, attach_perms=True).items()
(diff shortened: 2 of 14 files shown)


jonc125:
We could cache the list of ephemeral files names as per the patch below, but actually I'm not sure we need to!
```patch
diff --git i/weblab/repocache/models.py w/weblab/repocache/models.py
index 8034ac9..d5d6776 100644
--- i/weblab/repocache/models.py
+++ w/weblab/repocache/models.py
@@ -104,6 +103,19 @@ class CachedEntityVersion(VisibilityModelMixin):
         )


+class CachedEntityVersionFile(models.Model):
+    """
+    Cache for an 'ephemeral' file associated with a commit after it is made.
+
+    These are used for things like the Markdown README.md for a protocol, or the
+    errors.txt file listing protocol parsing errors.
+
+    They are stored in the repository as git notes, with the name cached here.
+    """
+    version = models.ForeignKey(CachedEntityVersion, related_name='files')
+    name = models.CharField(max_length=30)
+
+    class Meta:
+        unique_together = ['version', 'name']
+
+
 class CachedEntityTag(models.Model):
     """
     Cache for a tag in an entity's repository```

jonc125:
I've tested d046ed4 locally on my vagrant box and it successfully pulls a readme.md from WL1 protocols, so think this is now ready for review as fixing #75 - even though when we do #48 the implementation will probably change to use a Celery task for parsing.

Should merge after #97.

PR #102, opened 2019-01-15
Ensure data deleted when experiment deleted

### LINKED ISSUE #66: Ensure deleting an entity deletes linked simulation results

helenst:
Ensure related experiments are deleted when an entity is deleted. (This was already the case, but here we add tests to check this). Delete the experiment data directory upon deletion.

Also added tests to ensure entity repo is deleted on entity deletion.

Fixes #66

### DIFF
--- a/weblab/conftest.py
+++ b/weblab/conftest.py
@@ -109,6 +109,20 @@ def queued_experiment(model_with_version, protocol_with_version):
 
 
 @pytest.fixture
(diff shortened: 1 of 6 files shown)


jonc125:
Can't see any issues from a quick code review. I'll try to test on a local deploy later, just in case.

jonc125:
I've now run this on my local deploy, and confirmed that the experiment results were deleted when I deleted the model.

PR #103, opened 2019-01-15
Re-enable "select all" etc. in compare view

### LINKED ISSUE #100: Select all / none doesn't seem to work
on e.g.
https://scrambler.cs.ox.ac.uk/entities/protocols/59/versions/v1/compare

Child of #68

helenst:
Fix the "select all" buttons on the compare view.

Fixes #100


jonc125:
I can however confirm from https://scrambler.cs.ox.ac.uk/entities/protocols/9/versions/v1/compare that the 'show all versions' button does work, as does 'select latest'

helenst:
OK... that should be working now.

PR #109, opened 2019-01-24
Check protocol syntax on upload

### LINKED ISSUE #48: Implement protocol file syntax checking

jonc125:
Fixes #48, and extracts the protocol 'interface' in the process.

It doesn't give any error if there isn't a main file marked, just doesn't do anything. And the readme extracting is still done separately on the Django side (but this will probably remain the case until we update the runner to Python 3).

The testing possibly still needs to be beefed up, but I think this is probably at a stage where it's useful to review. I've confirmed that when deployed locally the check task does run as expected, including for error cases. Still need to update the automatic deployment though so it's easy for others to do this; will hopefully get to that later today/tomorrow.


helenst:
I haven't tried running it all the way through but the code looks good. Will require a rebase.

helenst:
OK, I've tested this out locally (well, with a dummy webservice and simulating the callback) and seems fine!

PR #111, opened 2019-01-30
Entity comparisons

### LINKED ISSUE #42: Add model comparison functionality
I.e. the ability to compare versions of the same entity (or indeed versions of different entities, including protocols not just models).

Mostly this is porting over views & plugins from the old Web Lab: BiVeS, diff, etc.

We may want to think about a UI for more easily comparing across entities (e.g. to support forks). But that's a separate issue.

Also note per https://bitbucket.org/joncooper/fcweb/commits/c418bde6fea976e4dcf44f7c81056d193bef1976 that the BiVeS URL has changed. Martin says:
> The instance at http://bives.sems.uni-rostock.de/ will stay alive for some time (proxying for https://bives.bio.informatik.uni-rostock.de/), but it may go down in the future, while https://bives.bio.informatik.uni-rostock.de/ will stay alive for ... longer... ;-)

helenst:
Fixes #42 
Hook into existing javascript and provide Bives and Unix diff endpoints and comparison pages.

I've also reworked the entity URL matchers to avoid duplication of patterns, and done a little refactoring around visibility logic. (And renamed existing comparisons of entities related to experiments, to avoid confusion).


weblab/entities/templatetags/entities.py
```diff
@@ -85,33 +118,33 @@ def url_version(entity, commit):
     a tag, or the tag contains a /, or the tag is one of our reserved
     names (new, latest), we fall back to the SHA1.
     """
-    url_name = 'entities:{}_version'.format(entity.entity_type)
+    url_name = 'entities:version'
```
jonc125: Unused variable

weblab/entities/tests/test_views.py
```diff
@@ -416,6 +417,175 @@ def test_returns_404_if_commit_not_found(self, client, logged_in_user):
...
+        data = json.loads(response.content.decode())
+        versions = data['getEntityInfos']['entities']
+        assert versions[0]['id'] == v1.hexsha
+        assert versions[0]['numFiles'] == 1
+        file_ = versions[0]['files'][0]
+        assert file_['id'] == 'file1.txt'
+        assert file_['name'] == 'file1.txt'
+        assert file_['author'] == model.author.full_name
+        assert file_['filetype'] == 'TXTPROTOCOL'
+        #assert file_['masterFile']
```
jonc125: Why is this commented out?

weblab/static/js/compare.js
```diff
@@ -595,29 +595,17 @@ function parseUrl (event)
...
-			basicurl = t.slice (0, i + 3).join ("/") + "/";
-			entityType = "protocol";
-			entityIds = t.slice (i + 3);
-		}
-    */
     if (parts[i] == 'experiments') {
       basicurl = parts.slice(0, i+2).join('/') + '/';
       entityType = 'experiment';
       entityIds = parts.slice(i+2);
+    } else if (parts[i] == 'models' && parts[i+1] == 'compare') {
```
jonc125: What about comparing two versions of a protocol? BiVeS doesn't make much sense in this case, but unix diff would. (Note that I don't think it's worth restricting what we offer depending on entity type: offer both, and users can figure out which is helpful for them!)

weblab/templates/experiments/experimentversion_detail.html
```diff
@@ -53,9 +53,9 @@ <h2 id="entityversionname">
         {% endif %}
 
         <br/>Corresponding model:
-        <a href="{% url 'entities:model_version' experiment.model.pk experiment.model_version %}">{{ experiment.model.name }} @ {{ experiment.nice_model_version }}</a>
+        <a href="{% url 'entities:version' 'model' experiment.model.pk experiment.model_version %}">{{ experiment.model.name }} @ {{ experiment.nice_model_version }}</a>
           &amp; protocol:
-          <a href="{% url 'entities:protocol_version' experiment.protocol.pk experiment.protocol_version %}">{{ experiment.protocol.name }} @ {{ experiment.nice_protocol_version }}</a>
+          <a href="{% url 'entities:version' 'model' experiment.protocol.pk experiment.protocol_version %}">{{ experiment.protocol.name }} @ {{ experiment.nice_protocol_version }}</a>
```
jonc125: This should pass `'protocol'` not `'model'`

weblab/entities/views.py
```diff
@@ -43,30 +50,22 @@
 from .models import Entity, ModelEntity, ProtocolEntity
 
 
-class ModelEntityTypeMixin:
+class EntityTypeMixin:
     """
     Mixin for including in pages about `ModelEntity` objects
```
jonc125: This docstring is now outdated

weblab/entities/views.py
```diff
@@ -301,12 +273,99 @@ def get_context_data(self, **kwargs):
...
+    """
+    def _file_json(self, entity, commit, blob):
+        """
+        JSON for a single file in a version of the entity
 
-class ProtocolEntityVersionCompareView(ProtocolEntityTypeMixin, EntityVersionCompareView):
-    pass
+        :param entity: Entity object
+        :param commit: `Commit` object
+        :param archive_file: ArchiveFile object
```
jonc125: The parameter appears to be called `blob` instead?

weblab/entities/views.py
```diff
@@ -721,3 +744,96 @@ def get_context_data(self, **kwargs):
             kwargs['formset'] = self.get_formset()
         kwargs['type'] = self.object.entity_type
         return super().get_context_data(**kwargs)
+
+
+class EntityDiffView(View):
+    def _get_unix_diff(self, file1, file2):
```
jonc125: Would it be worth creating a new (low priority) issue to use https://docs.python.org/3/library/difflib.html for this instead do you think?

weblab/entities/views.py
```diff
@@ -721,3 +744,96 @@ def get_context_data(self, **kwargs):
...
+
+        if len(files) != 2:
+            result = {
+                'responseText': 'invalid request'
+            }
+        elif diff_type == 'unix':
+            result = self._get_unix_diff(*files)
+        elif diff_type == 'bives':
+            result = self._get_bives_diff(*files)
+
```
jonc125: We should add an `else` here providing an error message if `diff_type` has not been set correctly.

weblab/entities/tests/test_views.py
helenst: It should've been removed - I copy pasted it from another bit of code that did need to serve up that field, but it looks as if this one doesn't. I'll remove the line.

weblab/entities/views.py
helenst: I wanted to use that, but it doesn't seem to provide the same style of diff as the `diff` command (i.e. to match the output given by WL1), unless I missed something.

weblab/static/js/compare.js
helenst: Yes, the URLS etc. are set up to support protocols, so this bit of code can easily be adapted to do so.

helenst:
Ready for re-review @jonc125

jonc125:
The code changes look good. I've been playing with it deployed locally and most things work perfectly.

One thing to fix:
- [ ] If you don't select any versions to compare and then click the button (e.g. to visit http://localhost:8088/entities/protocols/compare) you get a 404 error - it'd be nicer to get a proper page telling you to select something :)

helenst:
Fixed :)

Issue #65, opened 2018-08-02
Add the model annotation tool


jonc125:
This is the `editMetadata` 'visualiser', e.g. https://travis.cs.ox.ac.uk/FunctionalCuration/model/Aslanidiatrialmodel2009/29/v9ih2/555/aslanidi_atrial_model_2009cellml/444325/editMetadata

Steps required:
- [x] Implement `/protocol/get_interfaces` API endpoint, which requires #48.
    - Get the interfaces for the (latest versions of) all protocols visible to the user. Done in #112.
- [x] Implement a 'create new version' view that:
    - [x] Takes a single file's data directly
    - [x] Checks that the base version is as expected, to guard against simultaneous edits
    - [x] Handle re-running experiments involving the old version
    - [x] The latter two features should also be included in the normal 'new version' page (see #117)
    - [x] On successful AJAX submission we should redirect to the version view page
- [x] Fix up img references etc.

PR #112, opened 2019-01-30
Provide a view to get protocol interfaces

jonc125:
This is the first part of the work on #65, but can sensibly be reviewed as a standalone chunk.

It builds on #109, hence the base branch selected.

I still need to check that it's running the number of DB queries I expect...

Child of #68.


jonc125:
Having fixed the debug logging on a dev deploy (oops!) I've now checked that fetching http://localhost:8088/entities/protocols/get_interfaces does indeed result in the correct output and only 2 DB queries being made, so think this is ready for review.

weblab/repocache/populate.py
```diff
@@ -13,25 +11,26 @@ def populate_entity_cache(entity):
...
-        version = CachedEntityVersion.objects.create(
+        version = CachedEntityVersion.objects.get_or_create(
             entity=cached,
             sha=commit.hexsha,
-            timestamp=commit.committed_at,
-            visibility=visibility or entity.DEFAULT_VISIBILITY
-        )
+            defaults={
+                'timestamp': commit.committed_at,
+                'visibility': visibility or entity.DEFAULT_VISIBILITY,
```
helenst: This doesn't seem like it'll update the cache if the visibility of a version changes?

jonc125: Good spot, yes. I'll add a test case for that!

jonc125:
Fixed the issue you spotted and the merge conflict; now ready for re-review @helenst .

helenst:
Looks good to me :+1:

PR #114, opened 2019-02-15
Re-run expt support when adding a new entity version

jonc125:
Part of #65. Builds on #112. Child of #68.

- [x] When a new entity version is created, we need to add to a new DB table a list of the associated experiments to run (by listing model, model_version, etc.)
    1. Find parent commit sha(s)
    2. Look up `Experiment.objects.filter(**kwargs)` to get previously run experiments
    3. Record only those that the user can see
- [x] `EntityVersionJsonView` should provide the appropriate other entity+version from this table as an array
- [x] The `entity.js` code (around line 744) should use this array in a callback which calls itself at 10ms intervals, each time submitting a new experiment from the array, using code like in `db.js` line 18
- [x] `NewExperimentView` should remove any matching entry from the table if success
- [x] Don't try to submit pending experiments if you don't have permission to launch them


jonc125:
I've now verified this works when all the pieces are put together on a local vagrant deploy.

weblab/static/js/entity.js
```diff
@@ -750,6 +751,40 @@ function init() {
...
+   *     specifying the experiments to launch. May be empty.
+   * @param submitNow  whether to submit a job now, or just register the first callback.
+   */
+  function triggerExperimentSubmission(plannedExperiments, submitNow)
+  {
+    if (plannedExperiments && submitNow)
+    {
+      // Submit the first planned experiment to the queue
+      var exptSpec = plannedExperiments.pop();
+      exptSpec["task"] = "newExperiment";
```
helenst: This raises an error on the last iteration because `plannedExperiments` is an empty array so `pop()` returns undefined. The`if` statement needs a `.length` test on the array as boolean testing an array won't have the desired effect.

weblab/static/js/entity.js
```diff
@@ -750,6 +751,40 @@ function init() {
...
+    if (plannedExperiments)
+    {
+      // Call this function again after a delay
+      window.setTimeout(function(){triggerExperimentSubmission(plannedExperiments, true);}, 200);
+    }
+  }
+
   function deleteVersionCallback()
   {
     if (confirm("Are you sure to delete this version? (including all files and experiments associated to it)"))
```
helenst: Not part of the PR but just noticed the grammar is a bit odd on this message - should be "Are you sure you want to delete this version?" I think.

jonc125: Good spot, yes, will fix that.

weblab/static/js/entity.js
jonc125: Yeah, I fixed that in #116 when I noticed it. Should really pull the fix into this PR, but...

weblab/static/js/entity.js
jonc125: Though I don't think it's implemented yet anyway.

weblab/static/js/entity.js
helenst: Fair enough :) I'll take a look at that one in a minute.

PR #119, opened 2019-04-09
Cancel experiment runs when deleted

### LINKED ISSUE #106: Cancel running tasks when an experiment is deleted
Follows on from #102 . Child of #68 .

Otherwise if you delete an `ExperimentVersion` (possibly indirectly by deleting a model/protocol) while it is still in status queued or running, the run will still complete (tying up resources) then the results will be discarded.

Probably just needs a pre-delete hook for `RunningExperiment` to call the Celery cancel task.

jonc125:
Will fix #106.

Still need to:
- [x] test it really fixes the issue when deployed
- [x] don't call webservice if you're deleting a run that didn't start successfully!
- [x] fix tests: the patched `requests.post` doesn't cope with the cancel task call
- [x] add tests for new behaviour

### DIFF
--- a/weblab/experiments/apps.py
+++ b/weblab/experiments/apps.py
@@ -1,13 +1,14 @@
 from django.apps import AppConfig
 from django.db.models.signals import pre_delete
 
-from .signals import experiment_version_deleted
+from .signals import experiment_version_deleted, running_experiment_deleted
 
 
 class ExperimentsConfig(AppConfig):
     name = 'experiments'
 
     def ready(self):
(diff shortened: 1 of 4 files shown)


skeating:
Not sure how I can check that a run has been cancelled without access to the backend that is actually doing the running.

jonc125:
Thanks @skeating. It might be worth adding the monitoring commands you used to the deployment repo readme - there's a section there for 'Handy commands on the deployed server'.

Issue #64, opened 2018-08-02
Implement extra custom experiment matrix views


jonc125:
Replicate the WL1 buttons at the top of the matrix view for logged-in users, i.e.
* Show just moderated / all public / all visible-to-me / my experiments
* Extra buttons to hide moderated models / protocols from views that include these among others (all the above views include moderated entities by default)

The default view for anonymous users should be moderated experiments (which requires #41 & requires #40). The default for for logged-in users is 'my experiments' (which includes moderated ones as well by default).

helenst:
Tasks:

- [ ] Add "mine" filter to matrix json view
- [ ] Add "public" filter to matrix json view
- [ ] Add moderator visibility status
- [ ] Add "moderated" filter to matrix json
- [ ] Add moderator capability to users
- [ ] Ensure correct display of visibility dropdown depending on user moderator status
- [ ] Add hide/display moderated entities filter logic into matrix json view
- [ ] Add front end buttons and hook up to back end logic

jonc125:
Outstanding tasks not covered by #120:
- [ ] Correctly handle multiple versions with varying visibilities (see also #73)
- [ ] Set default view for logged-in users to 'my experiments'

PR #120, opened 2019-04-11
Experiment matrix filters

helenst:
For #64 

Includes the addition of "moderated" visibility status for entity versions.

There's a potential edge case around changing moderated visibility that I'm not sure about. If an entity is set as moderated and is viewed by a user who's not a moderator but has edit privileges on the entity, they don't get the moderated status in their dropdown, but still have the ability to set it to another status - i.e. remove it from the moderated set of entities. Is this something we're OK with? I think that is how WL1 currently deals with it.


weblab/static/js/db.js
```diff
@@ -499,42 +499,41 @@ function parseLocation ()
...
+          result[item[0]] = decodeURIComponent(item[1]);
+        });
       }
-      else if (items[0] == "all")
+      else if (items[0] == "moderated")
       {
-        $('#showAllExpts').addClass("selected");
-        ret.showAll = "1";
+        $('#showModeratedExpts').addClass("selected");
+        ret.subset = "moderated";
```
jonc125: Should this line also appear after line 542?

weblab/static/js/db.js
```diff
@@ -499,42 +499,41 @@ function parseLocation ()
...
+          }
+          if (item[0] == 'moderated-protocols' && item[1] == 'false') {
+            ret["moderated-protocols"] = "false";
+            $('#showMyExptsProtocols').text("Show moderated protocols");
+          }
+          result[item[0]] = decodeURIComponent(item[1]);
+        });
       }
-      else if (items[0] == "all")
+      else if (items[0] == "moderated")
```
jonc125: This should actually be the default if nothing is appended.

weblab/static/js/db.js
```diff
@@ -499,42 +499,41 @@ function parseLocation ()
...
       {
-        $('#showAllExpts').addClass("selected");
-        ret.showAll = "1";
+        $('#showModeratedExpts').addClass("selected");
+        ret.subset = "moderated";
       }
       else
```
jonc125: This shouldn't be the default case; instead it should be `if (items[0] == "all")`

weblab/experiments/urls.py
```diff
@@ -10,6 +10,8 @@
 urlpatterns = [
     url(
         '^'
+        '(?P<subset>mine|public|moderated)?'
```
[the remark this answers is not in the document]
jonc125: Actually, the options should be `mine|public|all` - moderated is the default view if nothing is appended.

weblab/experiments/views.py
```diff
@@ -84,9 +85,25 @@ def experiment_version_json(cls, version):
...
-        q_models = ModelEntity.objects.filter(id__in=visible_ids)
-        q_protocols = ProtocolEntity.objects.filter(id__in=visible_ids)
+        subset = request.GET.get('subset', 'visible')
+
+        if subset == 'visible':
+            entity_ids = visible_entity_ids(request.user)
+        elif subset == 'mine' and request.user.is_authenticated:
+            entity_ids = set(request.user.entity_set.values_list('id', flat=True))
+            if request.GET.get('moderated-models', 'true') == 'true':
+                entity_ids |= get_moderated_entity_ids('model')
```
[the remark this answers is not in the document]
helenst: I've not had time to update the logic (I suspect there will be more complexity than the entity query due to the way experiment visibilities are built up) but have corrected the way it works and included new test cases.

helenst:
I think this addresses all your comments @jonc125 - ready for re-review.

weblab/experiments/tests/test_views.py
```diff
@@ -248,59 +268,124 @@ def test_view_my_experiments_empty_for_anonymous(self, client, helpers, experime
...
-            experiment__protocol_version=my_protocol_public_version.hexsha,
+            experiment__protocol=other_protocol_moderated,
+            experiment__protocol_version=other_protocol_moderated_version.hexsha,
+        )
+
+        # Someone else's moderated model and public protocol: should not be visible
+        other_model_moderated = recipes.model.make(author=other_user)
+        other_model_moderated_version = helpers.add_version(other_model_moderated, visibility='moderated')
+
+        other_protocol_public = recipes.protocol.make(author=other_user)
```
jonc125: Similarly this was created above so can cut these 2 lines.

weblab/experiments/tests/test_views.py
```diff
@@ -248,59 +268,124 @@ def test_view_my_experiments_empty_for_anonymous(self, client, helpers, experime
         assert len(data['getMatrix']['experiments']) == 0
```
jonc125: Might be good if the test case had something to show here, actually, as a sanity check!

PR #128, opened 2019-04-13
Allow deleting (versions of) experiments

### LINKED ISSUE #108: Add views to delete (versions of) experiments
- Modelled on EntityDeleteView
- Both `Experiment[Version]` do have `is_deletable_by`
- So should just need to add appropriate url mappings, the views, links in the templates, and tests
- It might be nice for the button to be called 'cancel' rather than 'delete' if the experiment is queued or running.

Not sure if this should be considered part of #68 @mirams?

jonc125:
Fixes #108.

Tested deleting both a single version and an entire experiment on a local deploy, and both work.


skeating:
I get a FileNotFound error when trying to delete an experiment (whether it has multiple versions or not) . I get the same using master branch when trying Delete.

skeating:
Yes the two are definitely linked. If I have a model that has not run any protocols I dont get the error

jonc125:
Yeah, we haven't allowed deleting versions of models and protocols. This was allowed in WL1, but now they're backed by git repositories it would mean either re-writing the git history (ouch!) or adding some kind of extra metadata just to hide those versions. But you can effectively do the latter (at least for other users) by setting their visibility to private, so didn't seem worth adding.

jonc125:
> I think this may be the error that I fixed in #119 - [be3ffc7](https://github.com/ModellingWebLab/WebLab/commit/be3ffc72975d01a5b39722a4ff55269eb69afe93) - shall I cherry-pick that commit into this branch too?

I have no idea how I managed to make this comment 5 hours in the future!

MichaelClerx:
> I have no idea how I managed to make this comment 5 hours in the future!

Seeing this for other peoples commits too. Github was down a bit yesterday, must still be having issues

skeating:
Having found the correct view with icons I can confirm that you can delete versions of experiments - although presumably this is redundant once merged with [#118 Don't submit the same experiment twice](https://github.com/ModellingWebLab/WebLab/pull/118)

jonc125:
It's not redundant, since there are other reasons you might want to delete experiments. The main one would be tidying up instances that didn't work in order to free disk space. But also deleting a queued or running experiment would allow later submissions to run sooner.

skeating:
Yes I can see cancelling/deleting an experiment would be useful. - but if I'm interpreting it correctly #118 stops you creating two running instances (versions) of the experiment so you wont need to delete versions.

jonc125:
There are other valid ways in which multiple versions of an experiment can occur. You can explicitly re-run an experiment to create a new version. At present the main reason to do so is if the back-end simulation code has changed, since in theory all our simulations are deterministic. But we might support stochastic simulation in the future, and indeed fitting generally has an element of stochasticity to it.

jonc125:
Are you happy for me to merge this now @skeating ?

skeating:
@jonc125 yes merge away

PR #136, opened 2019-05-03
Tidy our HTML header template

jonc125:
Some things don't need to be ported from JSP as Django does things differently.

I haven't actually tested this, but am guessing it'll be ok ;)


skeating:
Seems to be fine. I couldn't spot any differences.

PR #138, opened 2019-05-09
Add view to run experiments based on a given entity version

### LINKED ISSUE #67: Add view to run experiments based on a given entity version
This is the 'Run experiments' button on an entity version view in WL1, e.g. https://travis.cs.ox.ac.uk/FunctionalCuration/batch/model/v9zjy/555. The corresponding JSP is https://github.com/ModellingWebLab/fcweb/blob/master/src/main/webapp/WEB-INF/Batch.jsp, which would need converting to a Django template. It uses some [Javascript code](https://github.com/ModellingWebLab/fcweb/blob/master/src/main/webapp/res/js/batch.js) for selecting/deselecting checkboxes and submitting the callback, but using jQuery properly the selection bit could be done in only a few lines! And the Django-y approach to submission would be an HTTP post rather than JS callback.

On submission we'd need to do something a bit like part of `entities.views.EntityNewVersionView` - record experiments to be run in the `PlannedExperiments` table (done by the `record_experiments_to_run` function there; logic would be slightly different here) and then either show the 'run experiments' page again or redirect to the entities:version page. I'm guessing the latter would be better.

We perhaps want to extend it by adding some filters to show _only_ my/my+moderated/my+public entities? Probably not for this issue though!

Low priority since you can run experiments via the matrix, it's just not as convenient for bulk operations, e.g. to easily re-run experiments when uploading a new entity version.

skeating:
Fixes #67.


weblab/entities/tests/test_views.py
```diff
@@ -1958,3 +1958,36 @@ def test_nonexistent_entity_redirects_anonymous_to_login(self, client, helpers,
     def test_nonexistent_entity_generates_404_for_user(self, client, logged_in_user, helpers, recipe, url):
         response = client.get(url % 10000)
         assert response.status_code == 404
+
+
+@pytest.mark.django_db
+class TestEntityRunExperiment:
```
jonc125: Tests look good!

weblab/entities/views.py
```diff
@@ -1013,3 +1013,43 @@ def get(self, request, *args, **kwargs):
         return JsonResponse({
             task: result
         })
+
+
+class EntityRunExperimentView(LoginRequiredMixin, EntityTypeMixin, DetailView):
```
jonc125: We probably also want this to inherit from `PermissionRequiredMixin` checking if the user can run experiments:
```python
    permission_required = 'experiments.create_experiment'
```

weblab/entities/views.py
```diff
@@ -1013,3 +1013,43 @@ def get(self, request, *args, **kwargs):
...
+        entity = self.object
+
+        context = super().get_context_data(**kwargs)
+
+        # preposition to use in sentence: You may run this entity on/under the following entities
+        context['preposition'] = 'under'
+        if entity.entity_type == 'protocol':
+            context['preposition'] = 'on'
+
+        # ended up using a nested dict as nessed lists caused django's unpacking in forloops to
```
jonc125: s/nessed/nested/

weblab/entities/views.py
```diff
@@ -1013,3 +1013,43 @@ def get(self, request, *args, **kwargs):
...
+        # mess things up slightly
+        other_entities = Entity.objects.filter(entity_type=entity.other_type)
```
jonc125: To improve query speed I think you could append `.select_related('cachedentity').prefetch_related('cachedentity__versions', 'cachedentity__versions__tags')`

weblab/entities/views.py
```diff
@@ -1013,3 +1013,43 @@ def get(self, request, *args, **kwargs):
...
+        context['object_list'] = []
+        for item in other_entities:
+            versions = item.cachedentity.versions
+            version_info = []
+            for version in versions.prefetch_related('tags'):
+                tag_list = list(version.tags.values_list('tag', flat=True))
+                commit = item.repo.get_commit(version.sha)
```
jonc125: If real-world performance turns out to be an issue we can probably do some clever tricks with repocache here, but not worth worrying about for now.

weblab/static/js/run_experiment.js
```diff
@@ -0,0 +1,51 @@
+
+var RunExperiment = function() {};
+
+RunExperiment.prototype = {
+  init: function() {
+    $("#checkallbutton").click (function () {
+        $(".latestexperimentCheckBox").each (function () {
+            if (!$(this).prop('checked')) {
```
jonc125: I think you can test with `if ($(this).prop('checked'))` and set with `$(this).prop('checked', true);` - that might get it working?

Also, if you're just setting not toggling, it's probably not worth the `if` guard.

weblab/templates/entities/entity_runexperiments.html
```diff
@@ -0,0 +1,75 @@
...
+{% block title %}Experiment: run specific combinations - {% endblock title %}
+
+{% block body_id %}runexperiment{% endblock %}
+
+{% block content %}
+  <h1 id="entityname">Run experiments using <span>{{ entity.name }}</span></h1>
+    <!-- SK TO DO: make these work -->
+     <p>
+         <input class="overrideCheckBox" type="checkbox" value="false"/>
+          Force overwrite of existing experiments
```
jonc125: Actually, this could probably be worded better - it won't actually overwrite, but will create a new version. Perhaps 'Create new versions of existing experiments (if unchecked, existing combinations will be skipped)'?

weblab/templates/entities/entity_runexperiments.html
```diff
@@ -0,0 +1,75 @@
...
+
+
+    <!-- SK TO DO: make these work -->
+     <button id="checkallbutton">check all</button>
+    <button id="checklatestbutton">check latest</button>
+    <button id="uncheckallbutton">uncheck all</button>
+   <p>
+        <form action="" method="post">
+      {% csrf_token %}
+       <button id="savebutton">run experiments</button>
```
jonc125: We might want to make this a normal Django form submission rather than an AJAX callback; not sure? Keeping closer to what WL1 did might be easier! Up to you :)

weblab/templates/entities/entity_runexperiments.html
```diff
@@ -0,0 +1,75 @@
...
+  <p>You may run this {{  type }} {{ preposition }} the following {{ other_type }}s.</p>
+
+  <h2> Your {{ other_type }}s</h2>
+
+    <!-- list each entity and all versions of that entity -->
+  {% for entity_object in object_list %}
+
+      <strong>{{ entity_object.id }}</strong>
+
+      {% for entity_version in entity_object.versions %}
```
jonc125: Should the loop be after the `<ul>`?

weblab/templates/entities/entity_version.html
```diff
@@ -40,12 +41,17 @@ <h2 id="entityversionname">
             <a href="{{ entity|url_tag_version:version }}">Add tag</a>.
             <form method="POST" action="">
               Change visibility: {{ form.visibility }}
-              <img src="{% static 'img/info.png' %}" alt="help" title="{{ entity.VISIBILITY_HELP }}" />.
+              <img src="{% static 'img/info.png' %}" alt="help" title="{{ entity.VISIBILITY_HELP }}" />
               <span id="versionVisibilityAction"></span>
-            </form>
-          {% else %}
+              <!-- SK: put in permission -->
```
jonc125: Yeah, I think you want this bit guarded only by `{% if perms.create_experiment %}` rather than whether they can modify the entity itself.

weblab/templates/entities/entity_version.html
```diff
@@ -40,12 +41,17 @@ <h2 id="entityversionname">
...
+              Run experiments:
+                <a title="Run experiments" href="{{ entity|url_runexperiments }}">
+                    <img src="{% static 'img/batch.png' %}" alt="Run experiments:" title="Run experiments using this model"/>
```
jonc125: `s/model/{{ entity.entity_type }}/`

weblab/entities/tests/test_views.py
skeating: Well Raquel did put in my aims for probation 'increase test coverage on a project by 75%' hard to do with weblab but I'm all for TDD

weblab/entities/views.py
skeating: fixed

weblab/static/js/run_experiment.js
skeating: Yeay it worked thanks you

weblab/entities/views.py
skeating: Done now fixing tests !!

weblab/templates/entities/entity_runexperiments.html
skeating: I was modelling this on what I found many other views did. I haven't come across a button that does something more than redirect that does not use this form; but I may have missed one :-)

weblab/templates/entities/entity_runexperiments.html
skeating: Fixed

weblab/entities/views.py
skeating: Done

weblab/templates/entities/entity_runexperiments.html
skeating: That looked okay; so have changed it as above

weblab/templates/entities/entity_version.html
skeating: Done

weblab/templates/entities/entity_version.html
skeating: Sorry not sure what you mean ??

jonc125: You have the `title` as "Run experiments using this model" - instead of model you should insert either model or protocol depending on the entity_type.

weblab/templates/entities/entity_runexperiments.html
jonc125: Indeed, I think this is the typical form structure for an AJAX submission. Probably like the `saveNewVersion` function in editMetadata.js; I dare say there are others. I was just wondering if it wanted to be more like `EntityNewVersionView` with a standard HTML form - but that would require more changes than matching the WL1 approach, so probably not worth doing.

weblab/templates/entities/entity_version.html
skeating: Doh !

Sorted

weblab/templates/entities/entity_runexperiments.html
skeating: Let me stick with whats working for me at present and then if/when we get round to tackling Michael's many issues we can add some clean up.

jonc125:
When adding to `PlannedExperiments` you may want to merge in my submit-expt-improvements branch (#169), which makes these default to re-run (since that's what you want to happen for the case when the first submission fails to connect, and all use-cases for this functionality so far make sense with this). And just make sure that either
1. if re-run is not selected, don't add existing combinations to `PlannedExperiments` (but possibly add notifications for these informing the user of the links to find the existing results!)
2. if re-run is selected, just add everything

skeating:
I admit defeat. @jonc125 Can you see why the line

        experiments_to_run = request.POST.getlist('runexperimentlist[]')
in my EntityRunExperimentView post function is not picking up checkboxes from the entity_runexperiments.html:

                 <input class="latestexperimentCheckBox" type="checkbox"
                         value="{{ entity_object.id }}:{{ entity_version.commit.hexsha }}"
                         name="runexperimentlist[]"
                         checked/>
                {%  else  %}
                   <input class="experimentCheckBox" type="checkbox"
                          value="{{ entity_object.id }}:{{ entity_version.commit.hexsha }}"
                          name="runexperimentlist[]"/>

jonc125:
I'll take a look on the train this morning!

weblab/entities/views.py
```diff
@@ -1004,3 +1004,63 @@ def get(self, request, *args, **kwargs):
...
+                latest = item.repo.latest_commit
+                version_info.append({'commit': commit, 'tags': tag_list, 'latest': latest == commit})
+            context['object_list'].append({'id': item.name, 'versions': version_info})
+        return context
+
+    def post(self, request, *args, **kwargs):
+        experiments_to_run = request.POST.getlist('runexperimentlist[]')
```
jonc125: For debugging you could `print(request.POST)` to see what actually gets passed.

weblab/entities/views.py
```diff
@@ -1004,3 +1004,63 @@ def get(self, request, *args, **kwargs):
...
+        for version in experiments_to_run:
+            ident, sha = version.split(':')
+            entity = Entity.objects.get(pk=ident)
+            commit = entity.repo.get_commit(sha)
+            record_experiments_to_run(request.user, entity, commit)
```
jonc125: You won't be able to use `record_experiments_to_run` as-is, since it is specifically for running the equivalent experiments on a new entity version that were run on its parents. But an equivalent approach of adding rows to `PlannedExperiment` should work here. In fact, here it should be easier since the POST arguments should contain (almost?) all the info you need directly, rather than needing to look at prior experiments - except for handling the re-run issue.

weblab/templates/entities/entity_runexperiments.html
```diff
@@ -0,0 +1,80 @@
...
+     </p>
+  <div id="runexperiment">
+
+
+    <!-- SK TO DO: make these work -->
+     <button id="checkallbutton">check all</button>
+    <button id="checklatestbutton">check latest</button>
+    <button id="uncheckallbutton">uncheck all</button>
+   <p>
+        <form action="" method="post">
```
jonc125: Possibly you need this to go just under the heading, so all the `input` fields fall within it?

weblab/templates/entities/entity_runexperiments.html
```diff
@@ -0,0 +1,80 @@
...
+      {% csrf_token %}
+       <button id="savebutton">run experiments</button>
+
+  </form>
```
jonc125: Similarly this might need to go at the end?

skeating:
Note the html page has to be different from the original as if I put all the buttons inside the django post command they try and rerun experiments

MichaelClerx:
Ugh. I keep getting errors from the DB when I switch between branches to try things out, even when I do only very normal operations (i.e. create a few models & experiments)

weblab/entities/templatetags/entities.py
```diff
@@ -189,3 +189,8 @@ def can_delete_entity(context, entity):
 def can_manage_entity(context, entity):
     user = context['user']
     return entity.is_managed_by(user)
+
+
+@register.filter
+def url_runexperiments(entity):
```
MichaelClerx: `url_run_experiments` ?

MichaelClerx:
I get errors occasionally, not sure how to recreate reproducibly:

```
Environment:


Request Method: GET
Request URL: http://localhost:8088/entities/models/45/versions/latest

Django Version: 1.11.20
Python Version: 3.5.2
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'social_django',
 'guardian',
 'accounts',
 'entities',
 'experiments',
 'repocache']
Installed Middleware:
['django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'rollbar.contrib.django.middleware.RollbarNotifierMiddleware']


Template error:
In template /opt/django/WebLab/weblab/templates/entities/entity_version.html, error at line 48
   Invalid filter: 'url_runexperiments'   38 :         by <em id="entityversionauthor">{{ version.author }}</em>.
   39 :         <div id="versionVisibility" data-change-href="{{ entity|url_change_version_visibility:version }}">
   40 :           {% if permission %}
   41 :             <a href="{{ entity|url_tag_version:version }}">Add tag</a>.
   42 :             <form method="POST" action="">
   43 :               Change visibility: {{ form.visibility }}
   44 :               <img src="{% static 'img/info.png' %}" alt="help" title="{{ entity.VISIBILITY_HELP }}" />
   45 :               <span id="versionVisibilityAction"></span>
   46 :                 {% if perms.create_experiment %}
   47 :                   Run experiments:
   48 :                     <a title="Run experiments" href=" {{ entity|url_runexperiments }} ">
   49 :                         <img src="{% static 'img/batch.png' %}" alt="Run experiments:"
   50 :                              title="Run experiments using this {{ entity.entity_type }}"/>
   51 :                 {% endif %}
   52 :            </form>
   53 :            {% else %}
   54 :             Visibility: {{ visibility }}
   55 :             <img src="{% static 'img/info.png' %}" alt="help" title="{{ entity.VISIBILITY_HELP }}" />
   56 :           {% endif %}
   57 :         </div>
   58 :       </small>


Traceback:

File "/opt/django/venv/lib/python3.5/site-packages/django/core/handlers/exception.py" in inner
  41.             response = get_response(request)

File "/opt/django/venv/lib/python3.5/site-packages/django/core/handlers/base.py" in _get_response
  217.                 response = self.process_exception_by_middleware(e, request)

File "/opt/django/venv/lib/python3.5/site-packages/django/core/handlers/base.py" in _get_response
  215.                 response = response.render()

File "/opt/django/venv/lib/python3.5/site-packages/django/template/response.py" in render
  107.             self.content = self.rendered_content

File "/opt/django/venv/lib/python3.5/site-packages/django/template/response.py" in rendered_content
  82.         template = self.resolve_template(self.template_name)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/response.py" in resolve_template
  64.             return select_template(template, using=self.using)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/loader.py" in select_template
  48.                 return engine.get_template(template_name)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/backends/django.py" in get_template
  39.             return Template(self.engine.get_template(template_name), self)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/engine.py" in get_template
  162.         template, origin = self.find_template(template_name)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/engine.py" in find_template
  136.                         name, template_dirs=dirs, skip=skip,

File "/opt/django/venv/lib/python3.5/site-packages/django/template/loaders/base.py" in get_template
  44.                     contents, origin, origin.template_name, self.engine,

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in __init__
  191.         self.nodelist = self.compile_nodelist()

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in compile_nodelist
  230.             return parser.parse()

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in parse
  515.                     raise self.error(token, e)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in parse
  513.                     compiled_result = compile_func(self, token)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/loader_tags.py" in do_extends
  312.     nodelist = parser.parse()

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in parse
  515.                     raise self.error(token, e)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in parse
  513.                     compiled_result = compile_func(self, token)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/loader_tags.py" in do_block
  255.     nodelist = parser.parse(('endblock',))

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in parse
  515.                     raise self.error(token, e)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in parse
  513.                     compiled_result = compile_func(self, token)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/defaulttags.py" in do_if
  969.     nodelist = parser.parse(('elif', 'else', 'endif'))

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in parse
  515.                     raise self.error(token, e)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in parse
  513.                     compiled_result = compile_func(self, token)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/defaulttags.py" in do_if
  969.     nodelist = parser.parse(('elif', 'else', 'endif'))

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in parse
  486.                     raise self.error(token, e)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in parse
  484.                     filter_expression = self.compile_filter(token.contents)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in compile_filter
  600.         return FilterExpression(token, self)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in __init__
  694.                 filter_func = parser.find_filter(filter_name)

File "/opt/django/venv/lib/python3.5/site-packages/django/template/base.py" in find_filter
  606.             raise TemplateSyntaxError("Invalid filter: '%s'" % filter_name)

Exception Type: TemplateSyntaxError at /entities/models/45/versions/latest
Exception Value: Invalid filter: 'url_runexperiments'
```

weblab/entities/templatetags/entities.py
skeating: Changed :-)

weblab/entities/views.py
```diff
@@ -1014,3 +1014,73 @@ def get(self, request, *args, **kwargs):
...
+        if not rerun or rerun == 'None':
+            exclude_existing = True
+        experiments_to_run = request.POST.getlist('runexperimentlist[]')
+        for version in experiments_to_run:
+            ident, sha = version.split(':')
+            exper_kwargs = {
+                this_entity.other_type + '_id': ident,
+                this_entity.other_type + '_version': sha,
+                this_entity.entity_type + '_id': this_entity.id,
+                this_entity.entity_type + '_version':  this_entity.repo.latest_commit.hexsha,
```
jonc125: Cache `this_entity.repo.latest_commit.hexsha` in a local variable outside the loop, otherwise you hit the filesystem every time around.

weblab/entities/views.py
```diff
@@ -1014,3 +1014,73 @@ def get(self, request, *args, **kwargs):
...
+            }
+            if exclude_existing:
+                existing = Experiment.objects.filter(**exper_kwargs)
```
jonc125: If you just want to check for existence it's slightly better to do:
```python
    if Experiment.objects.filter(**exper_kwargs).exists():
        continue
```

weblab/templates/entities/entity_version.html
```diff
@@ -40,12 +41,18 @@ <h2 id="entityversionname">
             <a href="{{ entity|url_tag_version:version }}">Add tag</a>.
             <form method="POST" action="">
               Change visibility: {{ form.visibility }}
-              <img src="{% static 'img/info.png' %}" alt="help" title="{{ entity.VISIBILITY_HELP }}" />.
+              <img src="{% static 'img/info.png' %}" alt="help" title="{{ entity.VISIBILITY_HELP }}" />
               <span id="versionVisibilityAction"></span>
-            </form>
-          {% else %}
+                {% if perms.create_experiment %}
```
jonc125: I think this `if` block needs to go outside the `if permission` block, so it's after line 56 below. You can run experiments using an entity even if you can't make changes to that entity.

weblab/entities/views.py
jonc125: Testing this on a local deployment I've discovered it skips experiments that have no versions, which isn't ideal! So perhaps instead we need
```python
    if ExperimentVersion.objects.filter(**exper_kwargs).exists():
```
and prepend `experiment__` to the `exper_kwargs` names.
```

File: weblab/entities/tests/test_views.py [hunk 20]
```diff
@@ -2054,3 +2054,351 @@ def test_nonexistent_entity_redirects_anonymous_to_login(self, client, helpers,
...
+        protocol = recipes.protocol.make(author=logged_in_user)
+        commit_protocol = helpers.add_version(protocol, visibility='public')
+
+        recipes.experiment_version.make(
+            status='SUCCESS',
+            experiment__model=model,
+            experiment__model_version=commit1.hexsha,
+            experiment__protocol=protocol,
+            experiment__protocol_version=commit_protocol.hexsha)
+
```
jonc125: To test behaviour with an experiment that has no versions, add:
```python
        recipes.experiment.make(
            model=model,
            model_version=commit2.hexsha,
            protocol=protocol,
            protocol_version=commit_protocol.hexsha)
```

weblab/entities/tests/test_views.py
```diff
@@ -2054,3 +2054,351 @@ def test_nonexistent_entity_redirects_anonymous_to_login(self, client, helpers,
...
+        response = client.get('/entities/models/%d/runexperiments/' % model.pk)
+        assert response.status_code == 200
+        assert response.context['object_list'] == [{'id': protocol.pk,
+                                                    'name': 'myprotocol1',
+                                                    'versions': [{'commit': commit2, 'tags': ['v1'], 'latest': True},
+                                                                 {'commit': commit1, 'tags': [], 'latest': False}]},
+                                                   ]
+        # Test post returns correct response
+        data = {'runexperimentlist[]': ['%d:%s' % (protocol.pk, commit2.hexsha)],
+                'entity.repo.latest_commit.hexsha': commit_model.hexsha,
```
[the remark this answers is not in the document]
skeating: I changed it to model_protocol_list since it is a list of model protocol version combinations to run. entity_commit didnt feel right

weblab/entities/tests/test_views.py
```diff
@@ -2054,3 +2054,351 @@ def test_nonexistent_entity_redirects_anonymous_to_login(self, client, helpers,
...
+        assert response.context['object_list'] == [{'id': protocol.pk,
+                                                    'name': 'myprotocol1',
+                                                    'versions': [{'commit': commit2, 'tags': ['v1'], 'latest': True},
+                                                                 {'commit': commit1, 'tags': [], 'latest': False}]},
+                                                   ]
+        # Test post returns correct response
+        data = {'runexperimentlist[]': ['%d:%s' % (protocol.pk, commit1.hexsha),
+                                        '%d:%s' % (protocol.pk, commit2.hexsha)],
+                'entity.repo.latest_commit.hexsha': commit_model.hexsha,
+                'rerun_expts': None}
```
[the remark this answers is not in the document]
skeating: added a seperate test for when no checkboxes are checked; there are tests for if the rerun_expts box is checked or not

weblab/entities/urls.py
```diff
@@ -135,4 +135,10 @@
         views.EntityDiffView.as_view(),
         name='diff',
     ),
+
+    url(
+        r'^%s/(?P<pk>\d+)/runexperiments/$' % _ENTITY_TYPE,
```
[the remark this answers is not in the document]
skeating: done

weblab/entities/views.py
```diff
@@ -1014,3 +1014,73 @@ def get(self, request, *args, **kwargs):
...
+        context = super().get_context_data(**kwargs)
+
+        # preposition to use in sentence: You may run this entity on/under the following entities
+        context['preposition'] = 'under'
+        if entity.entity_type == 'protocol':
+            context['preposition'] = 'on'
+
+        # ended up using a nested dict as nested lists caused django's unpacking in forloops to
+        # mess things up slightly
+        other_entities = Entity.objects.filter(entity_type=entity.other_type)\
```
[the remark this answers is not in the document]
skeating: changed and added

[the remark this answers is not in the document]
skeating: actually adding author makes all my tests fail - put this on hold for now

weblab/entities/views.py
```diff
@@ -1014,3 +1014,73 @@ def get(self, request, *args, **kwargs):
         return JsonResponse({
             task: result
         })
+
+
+class EntityRunExperimentView(PermissionRequiredMixin, LoginRequiredMixin, EntityTypeMixin, DetailView):
+    """
+    Class for listing the possible experiment combinations
```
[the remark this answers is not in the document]
skeating: done

weblab/entities/views.py
```diff
@@ -1014,3 +1014,73 @@ def get(self, request, *args, **kwargs):
...
+                context['object_list'].append({'id': item.id, 'name': item.name, 'versions': version_info})
+            else:
+                context['other_object_list'].append({'id': item.id, 'name': item.name, 'versions': version_info})
+        return context
+
+    def post(self, request, *args, **kwargs):
+        # this in not intuitive
+        # in get context self.object was the entity being worked with
+        # here we have to retrieve it
+        this_entity = self.get_object()
```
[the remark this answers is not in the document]
skeating: Yes although my issue was that having done been able to use self.object in the get_context method self no longer had an object when post was called - which is after the call to get_context !!

weblab/entities/views.py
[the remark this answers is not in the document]
jonc125: How odd! Indeed, not a priority.

weblab/entities/tests/test_views.py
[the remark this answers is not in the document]
jonc125: It was the next line I was commenting on, actually.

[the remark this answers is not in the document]
jonc125: But the renaming you did is also good :)

weblab/entities/views.py
```diff
@@ -1014,3 +1014,73 @@ def get(self, request, *args, **kwargs):
...
+
+        rerun = request.POST.get('rerun_expts')
+        exclude_existing = False
```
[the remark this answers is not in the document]
skeating: good one thanks

weblab/entities/views.py
skeating: fixed

weblab/entities/views.py
skeating: which doesnt work :-(

jonc125: That's bizarre. Are you doing things like
```python
            exper_kwargs = {
                'experiment__' + this_entity.other_type + '_id': ident,
```
?

skeating: aha is it 2 underscores ?

jonc125: Yeah, it's a double underscore - that's Django's way of describing lookups along foreign key relationships.

weblab/entities/views.py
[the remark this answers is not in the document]
skeating: Actually original is correct - we want any other other_entities which then get divided into 'My' and 'Other'

weblab/entities/views.py
skeating: done but requires a different set of kwargs - without the experiment__ to add to Planned Experiments

weblab/templates/entities/entity_version.html
skeating: Taken it out for now. Not sure where it does fit. I was going with the logic that by selecting combinations of model/protocols you were creating experiments and that if you did not have permission to do that there was no point in offering the option. I had thought we might need to filter the lists of offered combinations to ones that a user was allowed to 'create'  and 'run' - presumably only ones they already ran - if they dont have permission to run other peoples

weblab/entities/tests/test_views.py
skeating: added

jonc125:
This could do with a test that the view now works with a non-latest version of 'this_entity', but apart from that I think it's ready to merge.

jonc125:
Note there are some fixes for this now as part of https://github.com/ModellingWebLab/WebLab/pull/185, so we'll need to cherry-pick those into this branch before merging.

Also, given the speed of this view at the workshop was a little slow and the display was rather long, we might want to make a couple of extra changes.

skeating:
Okay I cannot change the code that loads the runexperiments view to not grab commits from the repo - and this will be heavy hitting as it queries each version of each entity ! 
See https://github.com/ModellingWebLab/WebLab/issues/191 for more detailed explanation.

I also cannot work out a way of fetching the commits without hitting the repo. I can get cachedentities with some of the information but I need the actual commit object for the hrefs call to url functions.

I accept I may be missing something but have spent some time playing with various queries etc and cannot see it.

skeating:
I would claim this done - again :-)

skeating:
Might really be done this time :-)

jonc125:
Yep :)

Issue #149, opened 2019-05-16
Version of experiment page is a bit unclear


MichaelClerx:
When I click the refresh button on a failed experiment, it's not terribly clear what's happening.

Also, leaving and returning to the page, I see the old error message. Hard to see if this is a the new version with the same error or the previous version.

Would probably be better to 

- Always jump to newest version, if user doesn't explicitly ask otherwise (even if the newest version hasn't completed yet)
- Add some red bar above the "Version 1" title, saying "You are viewing an older version of this experiment, to see the latest version click here"

jonc125:
I'm not sure about the idea of jumping to the latest version, but #170 adds a (blue) bar informing you when there's a newer version, and #169 makes the notification when launching an experiment contain a link to the new version. Do these do enough between them?

PR #170, opened 2019-05-23
Various navigation & UI improvements

### LINKED ISSUE #142: Markdown files not recognised as a type
We have these for lots of protocols, but user interface doesn't know about them?

Screenshot from 2019-05-16 11-24-02

### LINKED ISSUE #143: Change title "New version" to something more informative
e.g. model version, protocol version?

Technically creating a version, but user presumably will think in terms of creating a model? Button I just clicked was "create a new model"

### LINKED ISSUE #144: Capitalise "create new model/protocol"
Screenshot from 2019-05-16 11-30-29

### LINKED ISSUE #145: Can we get a "Create new model" on the experiments page?
The "Michael Clerx" button looks like I'd only use it for changing my profile picture, but is the core of the website...

### LINKED ISSUE #155: Versions are numbered? "Version: 1"
Bit unclear to me why I'm seeing "Version: 1" as the title of a version

For example:
- Start an experiment, it's called "Version 1"
- Click the "rerun an experiment" button
- Delete the experiment version
- List shows one version: "Version 1"

Easy to get confused!

### LINKED ISSUE #160: Relax constraints on allowed file names
```
the name 'data-sine-wave-cell-5.csv' contains reserved characters; only alpha-numeric characters, underscores and periods are allowed.
```

### LINKED ISSUE #165: Entity with zero versions should still have view
In bug situations (which will always occur) (e.g. #164 ) you can create an entity without versions. Clicking links for the entity takes you to a 404 page.
This should be replaced by some standard page instead that just says ```This [model/protocol/etc.] has no versions. (add new version)```

### LINKED ISSUE #166: CSV file type not recognised at upload
There is a CSV option in the dropdown, but it's not automatically selected

jonc125:
Fixes #145. Fixes #144. Fixes #143. Fixes #142. Fixes #166. Fixes #155. Fixes #165. Fixes #160.

Might also be enough for #149.


weblab/templates/includes/navbar.html
```diff
@@ -2,7 +2,11 @@
 <nav>
   <ul class="navbar">
     <li><a href="{% url 'home' %}">Home</a></li>
-    <li><a href="{% url 'experiments:list' %}">Experiments</a></li>
+    {% if user.is_authenticated %}
```
MichaelClerx: What's the rationale for this one?

jonc125: If the user is logged in, the link takes them to a view of primarily their experiments (+moderated). If not logged in, they just get moderated entities by default.

weblab/templates/entities/entity_newversion.html
```diff
@@ -6,7 +6,7 @@
 {% block content %}
   {% include "./includes/entity_header.html" %}
 
-  <h2>New version</h2>
+  <h2>Create a new version</h2>
```
MichaelClerx: Can we have the entity type in there as well?

jonc125: As in 'Create a new model version' or 'Create a new version of this model'?

weblab/experiments/models.py
```diff
@@ -159,7 +159,7 @@ class Meta:
 
     @property
     def name(self):
-        return str(self.run_number)
+        return '{:%Y-%m-%d %H:%M:%S}'.format(self.created_at)
```
MichaelClerx: We should probably have a shared constant somewhere that determines the output format for dates

weblab/static/js/upload.js
```diff
@@ -88,8 +88,8 @@ Upload.prototype = {
       error = "there is already a file with the name '" + name + "' - please remove that first.", "error";
     } else if (this.reserved_names.indexOf(name) != -1) {
       error = "the name '" + name + "' is reserved for system use; please choose another file name.", "error";
-    } else if (!/^[a-zA-Z0-9._]+$/.test(name)) {
-      error = "the name '" + name + "' contains reserved characters; only alpha-numeric characters, underscores and periods are allowed.", "error";
+    } else if (!/^[\w._: \-]+$/.test(name)) {
```
MichaelClerx: Is there nothing in Django or the standard library (e.g. urllib) we can use for this check instead of building our own regex?

MichaelClerx: Looking into this a bit I find additional checks as well, e.g. at most 244 characters for windows compatibility

weblab/templates/entities/entity_newversion.html
MichaelClerx: Yes please!

weblab/entities/urls.py
```diff
@@ -4,7 +4,7 @@
 
 
 _COMMIT = r'(?P<sha>[^^~:/ ]+)'
-_FILENAME = r'(?P<filename>[\w\-.]+)'
+_FILENAME = r'(?P<filename>[\w\-. \%:]+)'
```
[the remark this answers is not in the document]
jonc125: It was the JS regex that complained at you before.

weblab/experiments/models.py
jonc125: At present I think it's only used here and in the corresponding test. If it occurs more then yes.

weblab/static/js/upload.js
jonc125: It's a bit ugly because these things are being checked both by the JS code and Django. At least now those regexs are more similar to each other!

PR #180, opened 2019-05-29
Fix bug viewing version list

### LINKED ISSUE #178: Page not found - entity with newer private version
Steps to reprodiuce

1. Create a model X, make it public
2. Create a new version of X, make it private
3. Log in as a different user, X is visible in the public view
4. Click X, it shows you the old public version
5. Click "list all versions"

Page not found
Request url: http://localhost:8088/entities/models/3/versions/
Raised by: entities.views.EntityVersionListView

jonc125:
Fixes #178.

### DIFF
--- a/weblab/core/visibility.py
+++ b/weblab/core/visibility.py
@@ -26,6 +26,8 @@ LEVELS = [
     Visibility.MODERATED,
 ]
 
+VISIBILITY_LEVEL_MAP = {vis: LEVELS.index(vis) for vis in LEVELS}
+
 
 def get_joint_visibility(*visibilities):
     """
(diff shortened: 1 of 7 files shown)


MichaelClerx:
Getting a page not found error with this version when I click the (now private) model in the matrix --> Think it's sending me to the private version!

1. Create a model (moderated)
2. Add a new version (private)
3. Log in as a different user
4. Click the model in the experiment matrix

jonc125:
> Getting a page not found error with this version when I click the (now private) model in the matrix

I'm unable to reproduce this here. Did you load the matrix as a fresh page the second time, or did you get to it via the back button?

MichaelClerx:
I had to log out and then back in again, so fresh page!

jonc125:
I still can't reproduce the issue, so don't want it to hold up merging this pre-workshop. We can review if it still happens after that!

Issue #137, opened 2019-05-07
Implement 'fake' back-end fitting flow


jonc125:
Alongside preparing the front-end aspects, we can actually implement fitting in the back-end fc-runner, similarly to Aidan's prototype. The back-end web service can have an extra API added that takes a 'fitting spec' and 'dataset' as well as model and protocol, and runs a single hardcoded fit, returning results. A temporary hack in the front-end can then recognise a 'magic' protocol and call this new API instead of the standard simulation.

This will require changes to at least 4 repos, not just this one. But I've stuck the issue here for convenience! @MichaelClerx you can create separate issues if desired.

WebLab:
- [x] Recognise a protocol that has a .csv file in it, and call the new API rather than the standard experiment runner. Pass the protocol's URL as the dataset and fitting-spec parameters.
- [ ] Possibly the protocol should also contain a file with pints code to provide a realistic fitting spec?

fc-runner:
- [ ] Add the new API to `chastewebservice.py`.
    - [x] It should download and unpack the dataset and fitting spec, since from the above these should be real URLs and at least the dataset should contain a .csv with data.
- [x] Ensure it has suitable error handling logic as for the existing experiment runner (perhaps refactoring as needed)

weblab-fc:
- [ ] Implement the special-case fit in here? (Keeping with pycml etc for code generation.)

deployment (I can help with this bit @MichaelClerx - perhaps work on it together on Thurs?):
- [x] Deploy the new fc-runner (i.e. the branch of the deployment repo that tests all this must deploy the corresponding fc-runner branch)
- [x] Deploy the code that runs the fit (
- [x] Deploy the hacky WebLab branch

Related to #19 (if needed).

MichaelClerx:
Thanks @jonc125 . What does the current WL2 run on? Is it something we could swap out for the bitbucket code with Aidan's modifications?

jonc125:
Forward simulations for the current WL2 are still using the C++/Python hybrid from WL1. But fitting experiments can easily use something different. The fc-runner just treats the simulator as an executable anyway.

MichaelClerx:
Might be easiest to just stick Aidan's prototype in then, instead of redoing that in a branch of fc?

jonc125:
The idea was to move towards what we'll eventually be using for fitting, rather than just Aidan's code. Does that make sense? Can start with Aidan's code and then replace it though if that's easier?

MichaelClerx:
[...]

Issue #131, opened 2019-04-29
Create ExperimentalData app with basic model & views


jonc125:
This may still get split into sub-issues...

- [ ] Figure out what is common with Experiment (cf #130) and put in a mixin class / common 'dataset' app (see also https://github.com/ModellingWebLab/project_issues/wiki/Workshop-notes-2018)
- [ ] Need probably many-to-many links between `ExperimentalData` and `Protocol`, and only once a fit is performed do you tie down exactly which were used? @MichaelClerx ?
- Add views to (these are probably almost identical to the entity views):
    - [ ] User uploads a dataset
    - [ ] Show list of datasets owned by user
    - [ ] View a single dataset

MichaelClerx:
>  Need probably many-to-many links between ExperimentalData and Protocol, and only once a fit is performed do you tie down exactly which were used? @MichaelClerx ?

Not sure what you mean here! Would an `ExperimentalData` be a single time series? (Or a single matrix, more abstractly)?

jonc125:
`ExperimentalData` is the wet lab analogue of `Prediction` (currently called `Experiment`), i.e. a complete dataset with potentially multiple matrices (CSV files). So it's everything you need to compare your prediction against to do a complete fit.

Ideally I'm assuming you'd have one-to-many correspondence between a protocol version and experimental data (since you can repeat the experiment to get new datasets...) but realistically people are going to develop their own protocols to improve the match to the wet lab procedure, or clone existing ones, etc. Hence many-to-many.

Related to this, we need to think about versioning of datasets. Do we want ExperimentalData to have versions in the same way Prediction does (i.e. re-run exactly the same protocol on exactly the same model - results might still differ due to new code version or stochasticity)? Would versions differ only in metadata, or is changing files allowed? Or should new files require a new ExperimentalData object? (With a many-to-many mapping this isn't really a problem, providing you name things sensibly and/or have good search!)

mirams:
I think an update in the `Protocol` version that data is linked to should probably create a new version of the `ExperimentalData` too. To force re-runs etc. with the refined protocol description to better match how the experiments were done.

But probably need to stop and think for a little while whether we want:
 * one `FittingSpec` to work with one `ExperimentalData` (which may include many protocols); 
 * or whether we want one `FittingSpec` to work with multiple `ExperimentalData`s that each are linked to one `Protocol`.

[...]

PR #185, opened 2019-05-30
Hacky version of fitting spec support

### LINKED ISSUE #182: Hacky support for fitting specs
- [x] Add a boolean field to `ProtocolEntity` (well, will probably have to be on `Entity` actually) stating that this is a fitting specification, not (just) a forward simulation protocol.
- [x] If this field is set, pass extra 'data' and 'fittingspec' arguments when launching experiments using this protocol.
- [x] On the experiments view, add a button to toggle between 'Show parameter fitting experiments' / 'Show simulation experiments' (which will add an extra boolean `show_fits` to the JSON call)
- [ ] Allow owners to toggle it on/off? Initially just require it to be set on entity creation.

jonc125:
Will fix #182.

Also includes aspects of #137, #131.

It is hacky and completely untested! We should consider what needs to be done to tidy this up before merge, or possibly post-merge to enable cleaner review.

Some clear outstanding tasks, which may become separate PRs, in rough priority order:
- [ ] Add tests for the new features, especially around data upload & view.
- [ ] Fix any bugs revealed!
    - [ ] For some reason the dataset description is not being set
- [ ] Respect dataset visibility - e.g. at present anyone can overlay other users' private datasets. Possibly they can link their data to private protocols as well; I'm not sure.
- [ ] Link fitting specs to datasets (and via that to a protocol) so a 'fitting spec' protocol only contains the actual fitting specification. We definitely don't want to do that as part of this PR. It may get subsumed by longer term planning for how this will be architected, e.g. when `FittingSpec` becomes a concept in its own right.


jonc125:
Added another issue I spotted to the description.

skeating:
Currently the description is part of the file upload page. This is inconsistent with 'description' for other files upload which is on the creation page. It is probably an easier fix to move the field :-)

jonc125:
> Currently the description is part of the file upload page. This is inconsistent with 'description' for other files upload which is on the creation page. It is probably an easier fix to move the field :-)

Fine by me!

jonc125:
I'm merging this to master so it can be deployed on scrambler and stop it being broken!

PR #193, opened 2019-06-11
Adding tests for datasets

skeating:
Add tests for the additional code used to view datasets

- [x] models
- [x] urls
- [x] views
- [x] template tags

Untested files that do not yet have tests required

- [x] admin
- [x] apps
- [x] forms 
- [x] signals

Name changes

- [x] ExperimentalDataset => Dataset


skeating:
@jonc I would have expected a test for ExperimentalDatasetCreation to work in the same way the test for EntityCreation works. They both use a form and then direct to the addFiles page.

But for entity the following passes:
class TestEntityCreation:
    def test_create_model(self, logged_in_user, client, helpers):
        helpers.add_permission(logged_in_user, 'create_model')
        response = client.post('/entities/models/new', data={
            'name': 'mymodel',
            'visibility': 'private',
        })
        assert response.status_code == 302

        assert ModelEntity.objects.count() == 1

and for dataset the following passes:

class TestDatasetCreation:
    def test_create_model(self, logged_in_user, client, helpers):
        helpers.add_permission(logged_in_user, 'create_dataset', ExperimentalDataset)
        protocol = recipes.protocol.make(author=logged_in_user)
        response = client.post('/datasets/new', data={
            'name': 'mymodel',
            'visibility': 'private',
            'protocol': 'myprotocol'                  <== tried with string and actual protocol entity
        })

        assert ExperimentalDataset.objects.count() == 0

Does this indicate some deeply buried bug (it is nothing obvious) and should I keep digging or is it a consequence of the slight difference between fact  that an entity create might be a new version of something existing and a dataset is a one off creation. 
Thought I'd ask before letting the time spent go above 2 hours !

No stress on answer - I can do other tests until you have time.

skeating:
Apologies the above comment should have pinged @jonc125

jonc125:
Indeed, that does look wrong! It might be worth checking what the form sends normally (i.e. when a human submits it) as the value for each field. Possibly protocol should be the id? And the 200 response could be because it's sending an error page. You can print `response.content` iirc to see what the generated HTML looks like.

skeating:
@jonc125  Next query. 

I added tests for dataset file upload which mimic the tests for entity file upload.
But the line
        assert dataset.files.count() == 1 
hits an error as it wants an argument for count ?. The dataset.files object is an empty list.

What is more confusing is that the corresponding line in entities TestFileUpload
        assert model.files.count() == 1
the model.files object is a FileManager

jonc125:
I remember something like this confusing me. Let me check...

`Entity.files` is a reverse foreign key connection to `EntityFile`:
```python
class EntityFile(models.Model):
    entity = models.ForeignKey(Entity, related_name='files')
```
which is why you get a `...Manager` object - it allows you to query on the linked table.

The property is defined differently in `Dataset` however:
```python
    @property
    def files(self):
        if self.archive_path.exists():
            return ArchiveReader(str(self.archive_path)).files
        else:
            return []
```
so it is an actual list of `core.combine.ArchiveFile` objects.

Ah, I remember now. Because I wanted to enumerate actual files in a `Dataset` (as for `Experiment.files` I defined the file uploads with a different name:
```python
class DatasetFile(models.Model):
    dataset = models.ForeignKey(ExperimentalDataset, related_name='file_uploads')
```

So try `assert dataset.file_uploads.count() == 1`

Possibly we should rename the field in `EntityFile` for consistency!

skeating:
@jonc125 can I go ahead and use this branch to do name changes
ExperimentalDataset => Dataset

uniform fieldnames for everything ie entity.files => entity=>file_uploads

any others I encounter

jonc125:
Sure. Perhaps run suggestions past me first just in case :)

skeating:
> Sure. Perhaps run suggestions past me first just in case :)

No problem. I've added a list (of all 2 so far) to the main explanation of the PR and will consult you before adding others :-)

skeating:
@jonc125 If you have a moment to comment on how I could get my reverted commit to work everywhere not just locally it would be much appreciated

skeating:
Yes I did initially BUT the get url constructs a name of the archive file for a dataset as dataset_name.zip - which is the file it first looks for

skeating:
so I was trying to see if making the dataset name correspond works - it doesnt yet

jonc125:
> Yes I did initially BUT the get url constructs a name of the archive file for a dataset as dataset_name.zip - which is the file it first looks for

Ah, I see. Yeah, there should be separate tests for downloading the dataset as a whole archive, and for downloading an individual CSV file.

skeating:
Yes there will be - maybe I should do those first and then try individual files. - This is the never-ending PR :-)

skeating:
@jonc125 At some point could you have a look at the TestDatasetFileDownloadView tests.
I've added comments about where weirdness happens.
Even the commented out test in this class are strange so do have comments; but taken out of build for now

skeating:
Okay sending this your way - finally ! (I think this PR and my machine were cursed). There a couple of questions in my code about tests I tried that didn't work but wasnt sure whether they should or not. Sorry meant to do a PEP8 overview before I requested your review but would like this put to bed before I go away

jonc125:
Great, will look at this now. Don't worry about pep8 for this PR though; it can be done in #204.

weblab/datasets/urls.py
```diff
@@ -59,7 +59,7 @@
 
     # url(
     #     r'^(?P<pk>\d+)/delete$',
-    #     views.ExperimentalDatasetDeleteView.as_view(),
+    #     views.DatasetDeleteView.as_view(),
```
jonc125: Added a reminder of this in #131.

PR #200, opened 2019-08-05
199 upload datasets

skeating:
Fixes for uploading and displaying datasets

- [X] create a dataset fails with html error in dataset_files.html
- [x] create dataset does not actually upload file
- [x] problems when dataset file has same name as previously uploaded file

### DIFF
--- a/weblab/datasets/templatetags/datasets.py
+++ b/weblab/datasets/templatetags/datasets.py
@@ -1,8 +1,26 @@
+import math
+
 from django import template
 from django.core.urlresolvers import reverse
 
+from core.filetypes import get_file_type
+
 
 register = template.Library()
+
+
+@register.filter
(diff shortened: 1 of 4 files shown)


weblab/templates/datasets/includes/dataset_files.html
```diff
@@ -1,8 +1,8 @@
-{% load entities %}
+{% load datasets %}
```
jonc125: As discussed, save tidying this for #201.

weblab/datasets/templatetags/datasets.py
```diff
@@ -1,10 +1,28 @@
+import math
```
jonc125: Leave this kind of thing for #201.

skeating:
So in fact all my issues can from changing the id of the table

PR #204, opened 2019-08-14
Improve style checks and test them on travis

jonc125:
I've not fixed issues in the `datasets` app because there's no point until #193 is merged. This PR should then be updated and any remaining style issues fixed.

https://github.com/ModellingWebLab/project_issues/issues/43


jonc125:
@d20gdx this should be relatively straightforward to review - it's just adding the style check config and then fixing all the issues it threw up. Do ask if anything is unclear!

requirements/test.txt
```diff
@@ -12,14 +12,20 @@ defusedxml==0.5.0         # via python3-openid, social-auth-core
 dj-database-url==0.4.2
 django-braces==1.11.0
 django-guardian==1.4.9
-django==1.11.20           # via model-mommy
```
sroderick-g5sro: we've lost the "via model-mommy" comment. I assume thats irrelevant

setup.cfg
```diff
@@ -10,4 +10,4 @@ lines_after_imports=2
 order_by_type=1
 known_first_party=weblab
 include_trailing_comma=True
-skip=wsgi.py,migrations,__init__.py,node_modules
```
sroderick-g5sro: As an aside - I noticed that there is only a secrets.example.py file in the master not a secrets.py. I am assuming this is to force the installer to create their own copy.

setup.cfg
```diff
@@ -1,13 +1,29 @@
 [flake8]
-exclude=**/migrations/*,wsgi.py,__init__.py,**/node_modules/*
```
sroderick-g5sro: wsgi.py and __init__.py no longer being excluded,  is this correct?

setup.cfg
jonc125: That's right. No committing sensitive information to a public repo :)
(`secrets.py` is also in the .gitignore file to help avoid people adding it by accident)

requirements/test.txt
jonc125: Yeah, that's an artifact of an older pip-compile run I think. It's a direct dependency, and should always have been!

setup.cfg
jonc125: It is. I figured they didn't throw any style errors, so it didn't make sense to exclude them from the checks!

PR #205, opened 2019-08-16
Properly clean up storage folders

jonc125:
We spotted that test datasets weren't being cleaned up after the tests. This is due to the way pytest (and Django) handle database transactions. Each test is run in its own transaction which is rolled back at the end of the test so the DB is clean ready for the next test. But transaction rollback doesn't fire Django pre_delete signals. So WebLab never knew to delete repositories, dataset folders, etc. This issue was compounded by test datasets being created in the git clone, not a tmpdir, because we didn't realise that autouse fixtures were managing this.

So here we have a belt & braces fix. As well as creating an autouse fixture for the dataset storage location, the user fixtures always clean up after themselves for per-user disk storage. And the autouse fixtures have been tidied up a bit for consistency.

### DIFF
--- a/weblab/experiments/tests/test_views.py
+++ b/weblab/experiments/tests/test_views.py
@@ -1128,7 +1128,6 @@ class TestExperimentVersionJsonView:
         version.author.full_name = 'test user'
         version.author.save()
         version.status = 'SUCCESS'
-        # version.mkdir()
 
         response = client.get(
             ('/experiments/%d/versions/%d/files.json' % (version.experiment.pk, version.pk))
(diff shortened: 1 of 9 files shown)


weblab/experiments/tests/test_views.py
```diff
@@ -1128,7 +1128,7 @@ def test_experiment_json(self, client, logged_in_user, experiment_version):
         version.author.full_name = 'test user'
         version.author.save()
         version.status = 'SUCCESS'
-        version.abs_path.mkdir()
+        # version.mkdir()
```
skeating: did you mean to comment this out

skeating:
Lovely - does not leave multiple datasets hanging around.

weblab/experiments/tests/test_views.py
jonc125: Ah, no, it should be deleted! I was just commenting it out while testing to make sure it wasn't needed.

PR #211, opened 2019-09-10
Delete datasets

sroderick-g5sro:
please can you review.

### DIFF
--- a/weblab/datasets/tests/test_views.py
+++ b/weblab/datasets/tests/test_views.py
@@ -121,6 +121,87 @@ class TestDatasetCreation:
 
 
 @pytest.mark.django_db
+class TestDatasetDeletion:
+
+    def test_owner_can_delete_dataset(
+            self, helpers, logged_in_user, public_protocol , client
+    ):
+        helpers.add_permission(logged_in_user, 'create_dataset', Dataset)
+        dataset = recipes.dataset.make(author=logged_in_user, name='mydataset', protocol=public_protocol)
+        file_name = 'mydataset.csv'
+        file_contents = b'my test dataset'
+        recipes.dataset_file.make(  # Uploaded then deleted
+            dataset=dataset,
+            upload=SimpleUploadedFile(file_name, file_contents),
+            original_name=file_name,
+        )
+        client.post(
+            '/datasets/%d/addfiles' % dataset.pk,
+            data={
+                'filename[]': ['uploads/' + file_name],
+                'delete_filename[]': [],
+                'mainEntry': [file_name],
+            },
+        )
+        response = client.post('/datasets/%d/delete' % dataset.pk)
+        assert response.status_code == 302
+        assert response.url == '/datasets/'
+        assert not Dataset.objects.filter(pk=dataset.pk).exists()
+        assert not dataset.archive_path.exists()
+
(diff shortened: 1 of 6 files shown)


weblab/datasets/tests/test_views.py
```diff
@@ -120,6 +120,87 @@ def test_add_delete_add_file(self, logged_in_user, client, helpers, public_proto
...
+class TestDatasetDeletion:
+
+    def test_owner_can_delete_dataset(
+            self, helpers, logged_in_user, public_protocol , client
+    ):
+        helpers.add_permission(logged_in_user, 'create_dataset', Dataset)
+        dataset = recipes.dataset.make(author=logged_in_user, name='mydataset', protocol=public_protocol)
+        file_name = 'mydataset.csv'
+        file_contents = b'my test dataset'
+        recipes.dataset_file.make(  # Uploaded then deleted
```
jonc125: The comment here should probably be removed! And actually you could add `my_dataset_with_file` as an argument to the test (i.e. use that fixture) which lets you replace lines 129-145 by just `dataset = my_dataset_with_file`. And then you wouldn't need to use some other fixtures directly.

It might also be worth having a version of this test using `my_dataset` instead, to ensure users can delete 'partially created' datasets (i.e. without any files added yet).

weblab/datasets/tests/test_views.py
```diff
@@ -120,6 +120,87 @@ def test_add_delete_add_file(self, logged_in_user, client, helpers, public_proto
...
+        )
+        client.post(
+            '/datasets/%d/addfiles' % dataset.pk,
+            data={
+                'filename[]': ['uploads/' + file_name],
+                'delete_filename[]': [],
+                'mainEntry': [file_name],
+            },
+        )
+        response = client.post('/datasets/%d/delete' % dataset.pk)
```
jonc125: Perhaps add `assert dataset.archive_path.exists()` before this line to check it really accomplishes something?

weblab/datasets/tests/test_views.py
```diff
@@ -120,6 +120,87 @@ def test_add_delete_add_file(self, logged_in_user, client, helpers, public_proto
...
+        assert response.status_code == 302
+        assert response.url == '/datasets/'
+        assert not Dataset.objects.filter(pk=dataset.pk).exists()
+        assert not dataset.archive_path.exists()
+
+    def test_non_owner_cannot_delete_dataset(
```
jonc125: This one looks good!

weblab/datasets/tests/test_views.py
sroderick-g5sro: Done - thanks

weblab/datasets/tests/test_views.py
sroderick-g5sro: Yes - thanks

jonc125:
I still don't get why the PR build is failing on Travis though! But the branch build passes, so it might be an issue on master rather than this PR itself.

sroderick-g5sro:
Can this be merged or do I need to do anything else with it (e.g. around the failing checks)?

jonc125:
I think merge, as the branch itself builds fine.

PR #212, opened 2019-09-27
Added new tasks form for running and queued experiments

### LINKED ISSUE #207: List of queued or running experiments
Quite often I'll upload a model, start a batch of experiments, and then find an issue with the model.
This renders all running experiments with that version useless, as I'll have to add a new version which will replace the old version in the matrix.

It would be nice to have an option to:
- Kill all my simulations for a given model version (or at least unqueue queueueued ones)
- Kill all my simulations in general, ideally via a list of running and queued (but not finished!) experiments (a bit like windows' process manager)

This would save me a lot of time

sroderick-g5sro:
Fixes #207

I have created an initial write of the delete running experiments enhancement. I haven't finished the unit tests (because I need another day and also wanted to make sure I was on the right track first). I have used HTML forms rather than Django forms (forms.py) because it seemed easier to style these. I am also not sure about functionality for "admins" should they get a list of all running experiments. Currently I am only filtering on author. I haven't implemented a delete confirmation prompt or any send email functionality - can do if these are required.

### DIFF
--- a/weblab/experiments/tests/test_views.py
+++ b/weblab/experiments/tests/test_views.py
@@ -978,6 +978,7 @@ class TestExperimentTasks:
         running_exp = recipes.running_experiment.make(experiment_version=experiment_version)
         assert RunningExperiment.objects.count() == 1
         response = client.post('/experiments/tasks', {'chkBoxes[]': [running_exp.experiment_version.id]})
+        assert RunningExperiment.objects.count() == 1
         assert response.status_code == 404
 
 @pytest.mark.django_db
--- a/weblab/experiments/tests/test_views.py
+++ b/weblab/experiments/tests/test_views.py
@@ -21,7 +21,6 @@ from experiments.models import (
     PlannedExperiment,
     RunningExperiment,
 )
-from weblab.conftest import Helpers
 
 
 def generate_response(template='%s succ celery-task-id'):
@@ -902,7 +901,7 @@ class TestExperimentVersionView:
 class TestExperimentTasks:
 
     @pytest.mark.usefixtures('other_user')
-    def test_load_page_other_user(self,  client):
+    def test_load_page_other_user(self, client):
         response = client.get('/experiments/tasks')
         assert response.status_code == 302
 
@@ -915,12 +914,12 @@ class TestExperimentTasks:
     def test_get_queryset_other_user(self, other_user, client, experiment_version):
         experiment_version.author = other_user
         experiment_version.save()
-        running_exp = recipes.running_experiment.make(experiment_version=experiment_version)
+        recipes.running_experiment.make(experiment_version=experiment_version)
         assert RunningExperiment.objects.count() == 1
         response = client.get('/experiments/tasks')
         assert len(response.context['runningexperiment_list']) == 0
 
-    def test_get_queryset(self, logged_in_user, client,  helpers):
+    def test_get_queryset(self, logged_in_user, client, helpers):
         # Create three experiment versions 2 running and 1 completed
         model_1 = recipes.model.make(author=logged_in_user)
         model_1_version = helpers.add_version(model_1, visibility='public')
@@ -930,7 +929,7 @@ class TestExperimentTasks:
         protocol_2 = recipes.protocol.make(author=logged_in_user)
         protocol_2_version = helpers.add_version(protocol_2, visibility='public')
 
-        exp_version_1 = recipes.experiment_version.make(
+        recipes.experiment_version.make(
             status=ExperimentVersion.STATUS_SUCCESS,
             experiment__model=model_1,
             experiment__model_version=model_1_version.hexsha,
@@ -983,6 +982,7 @@ class TestExperimentTasks:
         response = client.post('/experiments/tasks', {'chkBoxes[]': [running_exp.experiment_version.id]})
         assert RunningExperiment.objects.count() == 1
         assert response.status_code == 404
+
 
 @pytest.mark.django_db
 class TestExperimentDeletion:
--- a/weblab/experiments/tests/test_views.py
+++ b/weblab/experiments/tests/test_views.py
@@ -900,7 +900,6 @@ class TestExperimentVersionView:
 @pytest.mark.django_db
 class TestExperimentTasks:
 
-    @pytest.mark.usefixtures('other_user')
     def test_load_page_other_user(self, client):
         response = client.get('/experiments/tasks')
         assert response.status_code == 302
@@ -939,7 +938,7 @@ class TestExperimentTasks:
         )
 
         exp_version_2 = recipes.experiment_version.make(
-            status=ExperimentVersion.STATUS_RUNNING,
+            status=ExperimentVersion.STATUS_QUEUED,
             experiment__model=model_1,
             experiment__model_version=model_1_version.hexsha,
             experiment__protocol=protocol_1,
@@ -963,13 +962,11 @@ class TestExperimentTasks:
 
         # Return only running experiment versions
         response = client.get('/experiments/tasks')
-        assert len(response.context['runningexperiment_list']) == 2
         assert set(response.context['runningexperiment_list']) == {running_exp_version2, running_exp_version3}
 
         # Cancel one of the running versions check the other one is still present
         client.post('/experiments/tasks', {'chkBoxes[]': [running_exp_version2.experiment_version.id]})
         response = client.get('/experiments/tasks')
-        assert len(response.context['runningexperiment_list']) == 1
         assert set(response.context['runningexperiment_list']) == {running_exp_version3}
         assert RunningExperiment.objects.count() == 1
 
--- a/weblab/experiments/tests/test_views.py
+++ b/weblab/experiments/tests/test_views.py
@@ -893,6 +893,21 @@ class TestExperimentVersionView:
 
 
 @pytest.mark.django_db
+class TestExperimentTasks:
+    def test_get_queryset(self, logged_in_user, client, queued_experiment):
+        response = client.get('/experiments/tasks')
+        assert response.status_code == 200
+
+    # def test_owner_can_delete_experiment(
+    #     self, logged_in_user, client, queued_experiment
+    # ):
+    #     assert queued_experiment.pk == 1
+    #     assert Experiment.objects.filter(pk=queued_experiment.pk).exists()
+    #     response = client.post('/experiments/task', {'chkBoxes[]': queued_experiment.pk})
+    #     assert Experiment.objects.filter(pk=queued_experiment.pk).exists()
+
+
+@pytest.mark.django_db
 class TestExperimentDeletion:
     def test_owner_can_delete_experiment(
         self, logged_in_user, client, experiment_with_result
--- a/weblab/experiments/tests/test_views.py
+++ b/weblab/experiments/tests/test_views.py
@@ -15,7 +15,10 @@ from django.test import Client
 from django.utils.dateparse import parse_datetime
 
 from core import recipes
-from experiments.models import Experiment, ExperimentVersion, PlannedExperiment
+from experiments.models import Experiment, ExperimentVersion, PlannedExperiment, RunningExperiment
+
+from weblab.conftest import Helpers
+
 
 
 def generate_response(template='%s succ celery-task-id'):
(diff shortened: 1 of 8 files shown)


sroderick-g5sro:
I wasn't sure how to distinquish between experiements run with different versions of models and protocols. For example... in the screen shot below:

image

both the "groupings" are for Model01/Protocol1 and they are only distinguished by their create date. In some cases this might be the same for different versions. Should I add a unique version identifier as is done below, on the models page

image

weblab/experiments/views.py
```diff
@@ -45,10 +47,33 @@ class ExperimentsView(TemplateView):
...
+    Show running and queued versions of all experiments
+    Delete checked versions
+    """
+    model = Experiment
+    template_name = "experiments/experiment_tasks.html"
+
+    def get_queryset(self):
+        return self.model.objects.filter(
+            Q(author=self.request.user) &
+            Q(versions__status=ExperimentVersion.STATUS_QUEUED) |
```
jonc125: Check precedence here - you might need extra brackets! (Probably better to have extra brackets for clarity anyway.)

weblab/experiments/views.py
```diff
@@ -391,6 +416,7 @@ def get_success_url(self, *args, **kwargs):
         return reverse('experiments:versions', args=[self.get_object().experiment.id])
 
 
+
```
jonc125: Try to avoid adding extra blank lines :)

weblab/static/js/experiment_tasks.js
```diff
@@ -0,0 +1,28 @@
+
```
jonc125: You'll need to add a `require()` line for this in the top of `main.js` to get it made available.

weblab/templates/experiments/experiment_tasks.html
```diff
@@ -0,0 +1,76 @@
...
+                            <div id="taskDisplay">
+                                <input class="taskCancelCheckBox" type="checkbox" name="chkBoxes[]"
+                                       value="{{ version.id }}">
+                                <div class="experiment-{{ version.status }}" id="taskDetailBox">
+                                    Version: {{ version.name }} by
+                                    <em>{{ version.author.full_name }}</em>
+                                    <div class="suppl" id="taskVersionDetailBox" style="padding-bottom: 0px">
+                                        <small>Created&nbsp;</small>
+                                        <time>{{ version.created_at }}&nbsp;</time>
+                                        {% with version.files|length as numfiles %}
```
jonc125: You probably don't need this level of detail!

weblab/templates/experiments/experiment_tasks.html
```diff
@@ -0,0 +1,76 @@
...
+            {% for experiment in object_list|dictsort:"name" %}
+                {{ experiment.name }}
+                <div class="suppl">
+                    <small>Experiment created
+                        <time>{{ experiment.created_at }}</time>
+                        by <em>{{ experiment.author.full_name }}</em>.
+                    </small>
+                </div>
+                <ul id="taskList">
+                    {% for version in experiment.versions.all|dictsortreversed:"created_at" %}
```
jonc125: Will need to check that previous completed versions of an experiment don't also show up here.

jonc125:
> I have used HTML forms rather than Django forms (forms.py) because it seemed easier to style these.

Looks good to me!

> I am also not sure about functionality for "admins" should they get a list of all running experiments. Currently I am only filtering on author.

I think that's fine. Admins should be able to delete via the auto-generated admin interface anyway.

> I haven't implemented a delete confirmation prompt or any send email functionality - can do if these are required.

I don't think so - @MichaelClerx ?

Regarding the display, `Experiment` has a `def get_name(self, model_version=False, proto_version=False)` function which you can use to add the further detail. Or you might need to create a property version `full_name` or similar; I can't recall offhand how easy it is for a template to pass arguments to a member function!

However, an alternative approach would be to restructure the queries so your model is `RunningExperiment` (for quicker queries), and use annotate to add details, e.g.
```python
return RunningExperiment.objects.filter(
    Q(experiment_version__author=self.request.user)
).order_by(
    'experiment_version__created_at',
).annotate(
    author_name=F('experiment_version__author__full_name'),
    expt_name=F('experiment_version__experiment__name'),
).select_related(
    'experiment_version',
)
```

MichaelClerx:
> > I haven't implemented a delete confirmation prompt or any send email functionality - can do if these are required.
>
> I don't think so - @MichaelClerx ?

Agreed!

weblab/experiments/views.py
sroderick-g5sro: as suggested changed query to use running experiments

weblab/experiments/views.py
sroderick-g5sro: noted - thanks

weblab/static/js/experiment_tasks.js
sroderick-g5sro: yes I have put this in now - thanks

weblab/templates/experiments/experiment_tasks.html
sroderick-g5sro: removed

weblab/templates/experiments/experiment_tasks.html
sroderick-g5sro: I have done some tests and it seems to filter by currently running experiment

sroderick-g5sro:
@jonc125 Hi, I have completed the changes to the code and hopefully addresses some of the review comments. I haven't really got any tests for this change., also the CI is complaining about Import sort orders. I wonder if you could approve the code changes Ive done please and I will sort out the rest on Monday 16th when I return from leave

weblab/templates/experiments/experiment_tasks.html
```diff
@@ -32,35 +32,18 @@ <h2>List of active tasks</h2>
...
-                                            <small>containing&nbsp;</small> {{ numfiles }} file{{ numfiles|pluralize }}
-                                        {% endwith %}
-                                    </div>
-                                </div>
+                    <li>
+                        <div id="taskDisplay">
+                            <input class="taskCancelCheckBox" type="checkbox" name="chkBoxes[]"
+                                   value="{{ running_experiment.experiment_version.id }}">
+                            <div class="experiment-{{ running_experiment.experiment_version.status }}" id="taskDetailBox">
+                               {{ running_experiment.experiment_version.experiment.name }} Version: {{ running_experiment.experiment_version.name }}
```
jonc125: As a minor efficiency improvement, you can add the `annotate` I suggested for `expt_name` and use `running_experiment.expt_name` rather than `running_experiment.experiment_version.experiment.name` - it'll save a DB query per experiment_version.

weblab/experiments/views.py
```diff
@@ -52,19 +49,21 @@ class ExperimentTasks(LoginRequiredMixin, ListView):
...
+        ).order_by(
+            'experiment_version__created_at',
+        ).select_related('experiment_version')
+        return q_running_experiments
 
     def post(self, request):
         """Process list of user selected checkboxes and kill running versions"""
         for ver in request.POST.getlist('chkBoxes[]'):
-            self.model.objects.filter(versions=ver).delete()
+            self.model.objects.filter(experiment_version_id=ver).delete()
```
jonc125: I think this now wouldn't work, as self.model is `RunningExperiment` but the ids you're getting back are the `ExperimentVersion` ids. So probably need `ExperimentVersion.objects.filter`?

weblab/static/js/main.js
```diff
@@ -10,6 +10,7 @@ var experiment = require('./experiment.js');
 var notifications = require('./lib/notifications.js');
 require('./compare.js');
 require('./entity_version_list.js');
+require('./experiment_tasks');
```
jonc125: I note all the others have `.js` appended - I'm presuming that this therefore isn't actually required?

weblab/templates/experiments/experiment_tasks.html
sroderick-g5sro: Thanks Jonathan. It didn't seem to let me return name using annotation. I got an error message with a list of supportable values none of which was name. I will try again though

jonc125: Ah, of course `Experiment.name` is a Python property not a DB field. Sorry!

Perhaps just `select_related` the experiment as well then?

weblab/experiments/views.py
sroderick-g5sro: thanks, It did seem to work ok.... Perhaps because the experiment version id field is also on the running experiemnts table? however your way is cleaner and more readable so I have implemented your suggestion here

weblab/static/js/main.js
sroderick-g5sro: yes, good spot thanks. I have corrected the extension. Not sure why it was working locally without that - does it perhaps use the local .js files? rather than the gulped one?

weblab/experiments/views.py
jonc125: Indeed, because all id fields are just sequences of integers, I suspect in the test both happened to be 1!

sroderick-g5sro:
@jonc125 Hopefully I have covered all the review comments now. Looks like I still need to address the sort imports causing travis still.

weblab/templates/experiments/experiment_tasks.html
sroderick-g5sro: now covered by unit tests - thanks

weblab/templates/experiments/experiment_tasks.html
sroderick-g5sro: yes thanks, I have added that

jonc125:
You can run `isort -rc .` on your machine to get isort to reorder imports automatically. The flake8 issues need to be addressed manually though.

weblab/experiments/tests/test_views.py
```diff
@@ -892,6 +897,93 @@ def test_view_experiment_version(self, client, experiment_version):
         assert response.context['version'] == experiment_version
 
 
+@pytest.mark.django_db
+class TestExperimentTasks:
+
+    @pytest.mark.usefixtures('other_user')
```
jonc125: I don't think we need the fixture here; the client uses an anonymous user by default unless the `logged_in_user` fixture is used.

weblab/experiments/tests/test_views.py
```diff
@@ -892,6 +897,93 @@ def test_view_experiment_version(self, client, experiment_version):
...
+            status=ExperimentVersion.STATUS_SUCCESS,
+            experiment__model=model_1,
+            experiment__model_version=model_1_version.hexsha,
+            experiment__protocol=protocol_1,
+            experiment__protocol_version=protocol_1_version.hexsha,
+            author=logged_in_user,
+        )
+
+        exp_version_2 = recipes.experiment_version.make(
+            status=ExperimentVersion.STATUS_RUNNING,
```
jonc125: Actually, maybe make this one QUEUED to prove that both statuses count?

weblab/experiments/tests/test_views.py
```diff
@@ -892,6 +897,93 @@ def test_view_experiment_version(self, client, experiment_version):
...
+            author=logged_in_user,
+        )
+        running_exp_version3 = recipes.running_experiment.make(experiment_version=exp_version_3)
+
+        assert ExperimentVersion.objects.count() == 3
+        assert RunningExperiment.objects.count() == 2
+
+        # Return only running experiment versions
+        response = client.get('/experiments/tasks')
+        assert len(response.context['runningexperiment_list']) == 2
```
jonc125: I wouldn't check both len and the set contents - checking the set contents will do both, and also give a more useful error message if the test fails (since you'll see which items didn't match).

weblab/experiments/tests/test_views.py
```diff
@@ -892,6 +897,93 @@ def test_view_experiment_version(self, client, experiment_version):
...
+        assert set(response.context['runningexperiment_list']) == {running_exp_version2, running_exp_version3}
+
+        # Cancel one of the running versions check the other one is still present
+        client.post('/experiments/tasks', {'chkBoxes[]': [running_exp_version2.experiment_version.id]})
+        response = client.get('/experiments/tasks')
+        assert len(response.context['runningexperiment_list']) == 1
```
jonc125: As above, delete this line.

weblab/experiments/views.py
```diff
@@ -45,6 +54,30 @@ class ExperimentsView(TemplateView):
...
+    """
+    Show running versions of all experiments
+    Delete checked versions
+    """
+    model = RunningExperiment
+    template_name = "experiments/experiment_tasks.html"
+
+    def get_queryset(self):
+        return RunningExperiment.objects.filter(
+            Q(experiment_version__author=self.request.user)
```
jonc125: You don't actually need the `Q()` wrapper for simple filtering cases like this; you can just do `.filter(experiment_version__author=self.request.user)`.

weblab/experiments/tests/test_views.py
sroderick-g5sro: thanks - changed

weblab/experiments/tests/test_views.py
sroderick-g5sro: good idea - done

weblab/experiments/tests/test_views.py
sroderick-g5sro: agreed

weblab/experiments/tests/test_views.py
sroderick-g5sro: agreed

weblab/experiments/views.py
sroderick-g5sro: good spot - done

PR #218, opened 2019-11-21
Use type-specific repocaches rather than a generic one for all entities

### LINKED ISSUE #217: Cache restructure
Based on the discussion in #203, restructure the `repocache` so we don’t need an `Entity` table any more. The plan is to have 3 sets of cache tables all sharing the same code, so we start talking about `CachedModel`, `CachedModelVersion`, `CachedModelTag`, etc.

Some changes are simple. For instance the `populate_entity_cache` command can iterate over `[ModelEntity, ProtocolEntity]` and call e.g. `populate_entity_cache(entity, CachedModel, CachedModelVersion, CachedModelTag)`. (We could store a map from the `ModelEntity` type to the relevant cache types, perhaps. Or use `CachedModel.entity.related_model` to go the other way, and the `CachedVersionClass` and `CachedTagClass` attributes to find the related caches.)

`entities.py` has two functions that operate over all entity types at present. Initially we can make these just query both model & protocol caches and combine the results. Later we may optimise away some use cases!
- `get_public_entity_ids` is used for the matrix view, and for `core.visibility.visible_entity_ids` to find entities visible to a given user. The latter is used for the matrix view, and in `ExperimentVersion.objects.visible_to` to filter for visible experiments in the compare view. That might not be the most sensible approach!
- `get_moderated_entity_ids` is used just for the matrix view, sometimes restricted to one type, sometimes both.

We should add the new DB model classes first, create the auto migration for that, then write a manual migration which copies the existing cache into the new tables, distributing based on `entity__entity_type` (with a reverse migration merging data). Then delete the old DB model classes and the auto-migration deleting the generic tables can happen.

The mixin classes implementing the functionality can assume the following properties defined by subclasses:
- `entity = models.OneToOneField(<EntityClass>, on_delete=models.CASCADE)`
    - Need to look up whether we can keep this in the base class; similarly for `CachedEntityVersion.entity`
- `EntityClass` - `ModelEntity` etc
- `CachedClass` - `CachedModel` etc
- `CachedVersionClass` - `CachedModelVersion` etc
- `CachedTagClass` - `CachedModelTag` etc
We can use the existing models’ code, appending `Mixin` to the class names, and removing their base classes (which will need to become bases for the concrete classes)? Or can we use
```python
    class Meta:
        abstract = True
```
And keep the base classes as they are? Which would be easier!

`ProtocolInterface` can just link to `CachedProtocolVersion`. Will need to see if the auto-migration copies data OK; I suspect not!

jonc125:
Fixes #217.

Done initially to see how hard it'd be, and turned out not too bad, though the migrations were a bit fiddly to get right!


jonc125:
Looks like the matrix view logic is broken when you compare results from multiple versions of an entity - it no longer excludes other entities. This wasn't picked up by tests, so we need to improve the tests and then fix this!

weblab/conftest.py
```diff
@@ -43,9 +42,9 @@ def add_version(entity,
         return commit
 
     @staticmethod
-    def add_fake_version(entity, visibility, date=None):
+    def add_fake_version(entity, visibility='private', date=None):
```
sroderick-g5sro: this is just a question for my own understanding pls... why are we changing this to be explicitly private here.

jonc125: This just sets a default visibility as private if none is supplied by the caller. I changed it to be consistent with `add_version`, so when moving tests to use 'fake' versions (i.e. purely in the DB, no git repository) I didn't need to add explicit visibilities when they were previously using the default.

sroderick-g5sro: thanks make sense. The change doesn't appear to be in the latest code for this branch, so I'm confused why its showing up in the diff

jonc125: It's in the last commit 224e49d I think, which is after I requested the review. So your local copy might be outdated?

sroderick-g5sro: yes that was the case. Pycharm wasn't doing what I expected. all good now. thanks

weblab/repocache/migrations/0012_split_cache.py
```diff
@@ -0,0 +1,117 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.20 on 2019-11-20 11:56
```
sroderick-g5sro: does this suggests it was auto generated? but we are adding manually?

jonc125: There's a Django manage.py command that creates the file template for you, filling in the migration dependencies.

weblab/repocache/models.py
```diff
@@ -124,6 +124,202 @@ class Meta:
         unique_together = ['entity', 'tag']
 
 
+####################################################################################################
+#
+# Mixin classes defining the common cache structure
+#
+
+class CachedEntityMixin(models.Model):
```
sroderick-g5sro: I'm going to need some pointers around these mixins please so I can appreciate what is going on. In the mean time I will keep digging

sroderick-g5sro: Ok., I think I get it now. Its just code common to the three child classes. sorry being thick

weblab/repocache/migrations/0012_split_cache.py
sroderick-g5sro: thanks.. I'm interested why we need the reverse case "combine_caches" do you have to support going back?

weblab/repocache/models.py
jonc125: That's right!

weblab/repocache/migrations/0012_split_cache.py
sroderick-g5sro: maybe we should call the file _split_and_combine_ cache

jonc125: yeah, mainly to support developers - so you can try out the branch, and roll back the DB if switching to another branch.

weblab/repocache/models.py
```diff
@@ -328,8 +328,7 @@ class ProtocolInterface(models.Model):
 
     A blank term is added to indicate that the interface has been analysed, in case of no actual terms being found.
     """
-    protocol_version = models.ForeignKey(CachedEntityVersion, on_delete=models.SET_NULL, related_name='interface', null=True, db_constraint=False)
-    new_protocol_version = models.ForeignKey(CachedProtocolVersion, on_delete=models.SET_NULL, related_name='interface', null=True, db_constraint=False, default=None)
+    protocol_version = models.ForeignKey(CachedProtocolVersion, on_delete=models.CASCADE, related_name='interface')
```
sroderick-g5sro: I could do with understanding why the changes have been made for on_delet and for removing db_constraint = false please.

weblab/repocache/migrations/0012_split_cache.py
jonc125: Regarding the rename query, I think most migrations I've seen have been named more like the forward direction, but it's not a major point.

weblab/repocache/models.py
jonc125: I think you're looking just at an intermediate commit there, right? So to get the migrations to operate without losing data, I had to split them into stages, because the `protocol_version` link was moving to point at a different table. So I had stage 1 adding a new field linking to the new table, making both links optional (`null=True`) and ensuring that if either source field was deleted the `ProtocolInterface` entry wouldn't be removed (`SET_NULL`). Then stage 2 gets rid of the old link, and stage 3 renames the new link to have the original name, making it required, and cascading deletions again. The `db_constraint` stuff was to stop the DB complaining during migrations IIRC.

sroderick-g5sro:
all looks fine to me., thanks

PR #219, opened 2019-11-25
Updated year in license

MichaelClerx:
Not sure I understand where the 2005 comes from!


jonc125:
2005 was the start of the Chaste project, which Web Lab built on (originally). Since we're using that CLA, we just kept the same boilerplate...

PR #221, opened 2019-11-28
Expand and refactor commit caching

### LINKED ISSUE #191: Unnecessary use of 'Commit' objects in views
There are several places in code where we use something like
                entity.repo.get_commit(version.sha))
 and pass the resulting commit object to an html view. Here it used to extract information about the commit - author/date/numfiles etc but there is no need for actual commit object. Since every commit records some information in the repocache surely it would be better to extend the information stored (some is there - some is not) so that there is no need to grab the actual commit.

What particularly confuses me is that the commit object gets passed to the url functions which seem to then create url based on tag if it exists or a reduced version of the sha. I haven't worked out the logic of this - but that also seems overkill for creating a url. The appropriate string is what is needed.

Note: I am talking only about views since these do not actually use anything other than attributes of the Commit. I realise other things may need the whole thing.

Disclaimer: I might be missing something crucial :-)

jonc125:
Part of #191.

I've added commit messages to the cache as a first example. Next steps:
1. Add/rename further columns in the cache as described in #191.
2. Replace uses of `Commit` objects by `CachedEntityVersion` objects wherever possible, especially in views.


weblab/conftest.py
```diff
@@ -26,13 +26,14 @@ def add_version(entity,
                     tag_name=None,
                     visibility=None,
                     cache=True,
+                    message='file',
```
sroderick-g5sro: would something like 'Add file' make more sense as a typical commit message. Or does the single term 'file' have signifiance I'm missing?

jonc125: It's only for tests, so doesn't really matter what it is. Some tests probably assume 'file' now.

weblab/repocache/migrations/0015_auto_20191128_1331.py
```diff
@@ -0,0 +1,35 @@
...
+            name='message',
+            field=models.TextField(default=' ', help_text='Git commit message'),
+        ),
+        migrations.AddField(
+            model_name='cachedprotocolversion',
+            name='message',
+            field=models.TextField(default=' ', help_text='Git commit message'),
+        ),
+        migrations.AlterField(
+            model_name='cachedmodelversion',
```
sroderick-g5sro: could you help me understand why there is an AlterField migration done to timestamp please

sroderick-g5sro: is it because of the addition of help text?

sroderick-g5sro: I've convinced myself that its because of the help text so obvious question - sorry

sroderick-g5sro:
Thanks Jonathan, well spotted with the missing references. Shall we close this PR and I'll start a new one for the next sub task.

jonc125:
Yep, sounds good. Since I created the PR I can't approve it, but take this as approval :)

Looks like you'll need to resolve some conflicts before merging it in though. Hopefully those will be straightforward.

jonc125:
It may be worth looking at some of the ideas in https://stackoverflow.com/questions/4300442/show-undefined-variable-errors-in-django-templates for future testing?

PR #235, opened 2020-01-15
Added support for caching of commit author

sroderick-g5sro:
Part of #191. This has been added as a database field as I believe we want to cache the committer of each version rather than simply returning the original author of the entity


weblab/repocache/models.py
```diff
@@ -80,6 +80,7 @@ def add_version(self, sha):
             sha=commit.sha,
             message=commit.message,
             master_filename=commit.master_filename,
+            author=commit.author.name,
```
jonc125: The docstring for `Commit.author` says it returns a string, so I'm not sure how the extra `.name` is working here! Though it appears to be from the tests below, so I'm confused.

But in any case, because we want the `author` field here to be the same as the `author` field on `Commit`, I don't think we want the `.name`. Does it break without?

jonc125: Or does `Commit.author` just need its docstring updating? It's possible that a GitPython `Actor` object renders the name when converted to string in templates?

Ah, or looking at the entity views, it might actually be using the wrong author in one place, and `author.name` in another. Ouch.

So I think your approach is correct after all.

PR #237, opened 2020-01-15
Added support for caching of numfiles

sroderick-g5sro:
Part of #191. Added support for the number of filename in each commit. It appears that just the number alone is currently required in the templates as opposed to the actual list


weblab/repocache/tests/test_populate.py
```diff
@@ -38,6 +38,7 @@ def test_populate(self, model_with_version, helpers):
         assert version2.master_filename == second_commit.master_filename == 'file1.txt'
         assert version2.author != version1.author
         assert version2.author == second_commit.author.name == 'another'
+        assert version2.numfiles == len(second_commit.filenames) == 2 # manifest.xml file has been added
```
jonc125: I'm guessing flake8 wants an extra space before the # !

PR #240, opened 2020-01-20
Swap call to repo to use enhanced repocache

### LINKED ISSUE #191: Unnecessary use of 'Commit' objects in views

sroderick-g5sro:
Fixes #191 

Changes to use enhanced repocache were possible.


sroderick-g5sro:
I have swapped obvious calls from repo to repocache. To proceed further,  I am going to need to cache the filenames and filesizes. @jonc125 is this your understanding? and shall I proceed to do that. I guess I will need new entities for these.

jonc125:
I'll try to take a look at this first thing tomorrow. Personally I don't think the gain from caching filenames is sufficient to be worth it at this stage.

weblab/repocache/models.py
```diff
@@ -67,6 +67,22 @@ def get_version(self, sha):
...
+    def get_name_for_version(self, sha):
+        """Get a human-friendly display name for the given version
+
+        :param sha: version sha
+        :return: first cached tag for this version, if any, or sha if not
+        """
+        try:
+            version = self.get_version(sha)
+        except ObjectDoesNotExist:
+            raise RepoCacheMiss("Entity version not found")
```
jonc125: `self.get_version` already raises `RepoCacheMiss` so you don't need the try-except here, just call the method.

(Hence why [coverage](https://codecov.io/gh/ModellingWebLab/WebLab/pull/240/changes#D1-70) shows an uncovered line.)

sroderick-g5sro: thanks all very helpful - appreciate it.

sroderick-g5sro: Its interesting.. get_commit() supportted getting a commit by tag name. I suppose I am going to have to support that in get_version() too.

jonc125: Yes - it's basically the converse of the get_nice_name, translating that name back into an actual commit.

sroderick-g5sro:
I'm just testing changes to EntityVersionMixin. I will commit these shortly.

weblab/repocache/models.py
sroderick-g5sro: The `string_if_invalid` works very well!! - 
image


image

sroderick-g5sro:
I hope this is now ready for re-review. In some of the tests I decided to stick with `add_version`  and the lookup the cached version later. An alterative approach I tried was to switch to `add_fake_version`. But this causes problems with add_tag. I tried to do an add_version_tag but the repo option has a tags or tag property that the repocache doesn't, this property is required by one of the mixins. Hopefully my approach is acceptable?

sroderick-g5sro:
I have had several attempts (see check ins) to add the `string_if_invalid` option to test.py but it has failed referencing BASE_DIR. I am not sure if there is an advantage of adding this in test. Possible for test systems such as Travis

weblab/config/settings/dev.py
```diff
@@ -80,3 +80,34 @@
...
+
+class InvalidStringShowWarning(str):
+    def __mod__(self, other):
+        import logging
+        logger = logging.getLogger(__name__)
+        logger.warning("In template, undefined variable or unknown value for: '%s'" % (other,))
+        return ""
+
+
+TEMPLATES = [
```
jonc125: Instead of putting the whole `TEMPLATES` definition here, can you just set the single relevant option? i.e.
```python
TEMPLATES[0]['OPTIONS']['string_if_invalid'] = InvalidStringShowWarning("%s")
```

weblab/config/settings/test.py
```diff
@@ -1,4 +1,15 @@
-from .base import *  # noqa
+from .base import TEMPLATES  # noqa
```
jonc125: You need the `*` here or most of the test settings won't be defined.

I don't know why the `RuntimeError` is happening though! Do you get any errors running `pytest` locally? Travis is just running `pytest --cov=weblab --cov-config=weblab/.coveragerc weblab`

weblab/entities/views.py
```diff
@@ -204,7 +234,7 @@ def get_context_data(self, **kwargs):
 class EntityVersionJsonView(EntityTypeMixin, EntityVersionMixin, SingleObjectMixin, View):
     def _planned_experiments(self):
         obj = self._get_object()
-        commit = self.get_commit()
+        commit = self.get_version()
```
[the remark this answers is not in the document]
sroderick-g5sro: ok thats good thinking.. thanks

weblab/config/settings/test.py
jonc125: Hmm, maybe we need to add
```
- export DJANGO_SETTINGS_MODULE="config.settings.test"
```
to the script section of .travis.yml ?

jonc125: (looking at https://pytest-django.readthedocs.io/en/latest/configuring_django.html)

jonc125: Ooh, and skimming the docs I've found https://pytest-django.readthedocs.io/en/latest/helpers.html#pytest.mark.ignore_template_errors. So in theory we can avoid touching `test.py` and instead change .travis.yml to call
```
pytest --fail-on-template-vars --cov=weblab --cov-config=weblab/.coveragerc weblab
```

weblab/entities/views.py
```diff
@@ -123,6 +123,36 @@ def get_commit(self):
...
+        """
+        Get the cached entity for this  version
+
+        :return: CachedEntityVersion object
+        :raise: Http404 if version not found
+        """
+        sha_or_tag = self.kwargs['sha']
+        cached_entity = self.object.cachedentity
+        try:
+            if sha_or_tag == 'latest':
```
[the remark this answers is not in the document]
sroderick-g5sro: The line `cached_entity.tags.get(tag=sha_or_tag)` returns a CachedEntityTag but we are expecting a CachedEntity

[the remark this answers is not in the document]
jonc125: Good point! You should be able to stick a `.version` on the end I think.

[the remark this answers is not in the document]
sroderick-g5sro: its ok, It just needed .version on the end.

weblab/entities/views.py
```diff
@@ -123,6 +123,36 @@ def get_commit(self):
...
         return self._commit
 
+    def get_version(self):
+        """
+        Get the cached entity for this  version
+
+        :return: CachedEntityVersion object
+        :raise: Http404 if version not found
+        """
+        sha_or_tag = self.kwargs['sha']
```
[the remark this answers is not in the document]
sroderick-g5sro: Yes I have been wrestling with this question of `_version`. I can see that here it is useful to protect against "requerying" when there is already a version. `_commit` comes into play because of the `__eq__ `override in Commit class. 

I have changed to using _version in get_version but haven't implemented the equality class (in CachedEntityVersion) and all the test pass. So hopefully this is going to be ok?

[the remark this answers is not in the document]
jonc125: Django has a default implementation of `__eq__` for DB model subclasses which checks that the IDs match. Which would do the right thing for us here, so should be OK.

weblab/config/settings/test.py
sroderick-g5sro: great thanks

sroderick-g5sro:
Some tests are failing now with missing filenames so looks like I have been over zealous with the swapping - on it now,.

jonc125:
Rather than back out most of the changes as you've done in f5c08eb, you can change the references in templates. The error looks like "Undefined template variable 'commit.filenames' in 'entities/entity_versions.html'". We didn't cache filenames, but we did I think cache numfiles, and most occurrences of filenames in templates pipe it through len, so the number of files is what they actually need, not the names. In other words, change `commit.filenames|length` to `commit.numfiles` in templates and we should be OK I think.

sroderick-g5sro:
ah yes.. sorry I must having been having a few funny mins. makes sense I will do that now.

sroderick-g5sro:
Made that change... I am now somewhat concern that I can't see my files in say the following screen...
image
I can see that we are using Javascript to display these. I will investigate if that code requires something that we are not passing and might be excluded from the `fail-on-template-vars` test

jonc125:
I'll run it here in a bit and check whether I see the same thing.

sroderick-g5sro:
thanks it works ok on linux. I'm just seeing if I can spot anything obvious on windows

jonc125:
I had to run `./manage.py populate_entity_cache` on my local dev setup to fill in the new cache entries (hence the little hack in b12368d!) but everything seems to be displaying correctly. I can also tell that views like run experiments load much faster, which is good!

It might just be that your browser needs to re-fetch the latest JS code into its cache?

Spotted one task remaining: `_url_friendly_label` in `entities/templatetags/entities.py` still uses the repo where it could use the cache. After that I think #191 is done!

sroderick-g5sro:
I've had a look at replacing the use of tag_dict as suggested. I have come up with the following:
```
def _url_friendly_label(entity, commit):
    """
    Get URL-friendly version label for a commit

    :param entity: Entity the commit belongs to
    :param commit: `git.Commit` object
    """
    tag = entity.cachedentity.versions.get(sha=commit.sha).tags.last()
    if tag is None or tag.tag in ['new', 'latest']:
        return commit.sha
    return tag.tag
```
 
unfortunately. It seems that sometimes the entity passed is a model but the commit was for a protocol.  I have gone back to master and this was still the case. It didn't matter because repo.tag_dict presumably has both type of commits.

I cant, at the moment. understand why the passed entity type is wrong. One approach I tried was to query all cached objects with CACHED_VERSION_TYPE_MAP but that seems overkill. 

I guess the other option is to maintain our own repocache.tag_dict

I wondered if you had any suggestions please? In the meantime I will try and understand why models are being passed in by the templates for a protocol commit.

jonc125:
That's indeed the replacement code I'd expect. The commit not belonging to the entity sounds like a bug! In such circumstances the original code will just return the commit.sha - the `entity.repo.tag_dict.get(commit.sha, ['/'])` bit returns the supplied default value `['/']`, rather than raising an error as the repocache version will do, so the bug has been hidden.

Are there particular tests that hit this bug? Or just in general browsing?

jonc125:
Possibly the culprit is
```diff
diff --git i/weblab/templates/entities/entity_runexperiments.html w/weblab/templates/entities/entity_runexperiments.html
index 223ae0d..f1f5505 100644
--- i/weblab/templates/entities/entity_runexperiments.html
+++ w/weblab/templates/entities/entity_runexperiments.html
@@ -46,7 +46,7 @@
                        name="model_protocol_list[]"/>
               {% endif %}
               <strong>
-                <a class="entityversionlink" href="{% entity_version_url 'version' entity entity_version.commit %}">
+                <a class="entityversionlink" href="{% entity_version_url 'version' entity_object entity_version.commit %}">
                   {% include "./includes/version_name.html" with tags=entity_version.tags version=entity_version.commit only %}
                 </a>
               </strong>
@@ -84,7 +84,7 @@
                        name="model_protocol_list[]"/>
               {% endif %}
               <strong>
-                <a class="entityversionlink" href="{% entity_version_url 'version' entity entity_version.commit %}">
+                <a class="entityversionlink" href="{% entity_version_url 'version' entity_object entity_version.commit %}">
                   {% include "./includes/version_name.html" with tags=entity_version.tags version=entity_version.commit only %}
                 </a>
               </strong>
```

weblab/entities/templatetags/entities.py
```diff
@@ -134,10 +134,14 @@ def _url_friendly_label(entity, commit):
     :param entity: Entity the commit belongs to
     :param commit: `git.Commit` object
```
jonc125: The docstring needs updating: we now need a `CachedEntityVersion` object here.

jonc125:
I've done a few experiments to see how much this PR improves the open file handles issue. I ran the entities test_views using the following patch to measure how many files are left open by the end. It only works on Mac or Linux though.
```diff
diff --git i/weblab/entities/tests/test_views.py w/weblab/entities/tests/test_views.py
index 8eefff7..c9ca969 100644
--- i/weblab/entities/tests/test_views.py
+++ w/weblab/entities/tests/test_views.py
@@ -23,6 +23,25 @@ from repocache.models import ProtocolInterface
 from repocache.populate import populate_entity_cache


+def get_open_fds():
+    '''
+    return the number of open file descriptors for current process
+
+    .. warning: will only work on UNIX-like os-es.
+    '''
+    import subprocess
+    import os
+
+    pid = os.getpid()
+    procs = subprocess.check_output(
+        ["lsof", '-w', '-Ff', "-p", str(pid)]
+    ).splitlines()
+    print(procs)
+
+    nprocs = len([s for s in procs if s and s.decode()[0] == 'f' and s.decode()[1:].isdigit()])
+    return nprocs
+
+
 @pytest.fixture
 def analysis_task(protocol_with_version):
     """A single AnalysisTask instance with associated Protocol version & repocache set up."""
@@ -2668,3 +2687,8 @@ class TestEntityRunExperiment:
             assert planned_experiment.protocol == protocol
             assert planned_experiment.protocol_version == proto_commit1.sha
             assert (planned_experiment.model, planned_experiment.model_version) in expected_model_versions
+
+
+def test_zzz_open_fds():
+    num_open_files = get_open_fds()
+    assert num_open_files < 3
```

* On master there were 81. With the change in 7e91e7c this went down to 57.
* On this branch there were 57. With 7e91e7c it went down to 39. With 6ce08ec it went down to 27.

So looking promising!

Issue #202, opened 2019-08-12
Respect dataset visibility


jonc125:
The initial implementation in #185 didn't include any visibility checks, so e.g. at present anyone can overlay other users' private datasets. (Because `experimentversion_detail.html` just iterates over `experiment.protocol.protocol_experimental_datasets.all`)

- [x] Ensure that a new dataset can't be linked to a non-visible protocol - probably requires filtering in the appropriate form? (cf https://github.com/ModellingWebLab/WebLab/pull/222/files#diff-d5c6e6ad5fedad21b166e67f131beb44)
    - [ ] Maybe even sort the visible protocol list, so your own protocols come first, followed by moderated ones? This isn't necessary though, so only if it's easy!
- [x] Ensure the only datasets you can overlay are ones you can see
    - [x] Filter the list presented, possibly by adding a helper `templatetag` that does the filtered DB query
    - [x] And ensure that if a non-visible id is passed to the view `datasets:version_json`, it gives an appropriate error
- [x] Make sure we have tests checking the above work

#184 might be a useful refactoring before tackling this issue? I don't think it's necessary though, so probably not.

jonc125:
This would be a reasonable issue to work on while waiting for my review @sroderick-g5sro.

PR #245, opened 2020-01-30
Respect dataset visibility

sroderick-g5sro:
Relates to issue #202

### DIFF
--- a/weblab/experiments/tests/test_templatetags.py
+++ b/weblab/experiments/tests/test_templatetags.py
@@ -16,14 +16,22 @@ def test_url_experiment_comparison_json():
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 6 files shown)


weblab/experiments/tests/test_templatetags.py
```diff
@@ -14,6 +15,17 @@ def test_url_experiment_comparison_json():
     assert exp_tags.url_experiment_comparison_json([]) == '/experiments/compare/info'
 
 
+@pytest.mark.django_db
+def test_dataset_options(logged_in_user, client):
```
jonc125: To make this a bit more robust, I'd use the `other_user` fixture as well, and create a private dataset owned by other_user linked to the same protocol, so we're certain private datasets won't be revealed by this query.

PR #246, opened 2020-01-31
Associate planned experiments with submitting user

### LINKED ISSUE #244: Run pending experiments only for user who owns experiments
Current situation (according to @jonc125 )

> This is a difference between WL1 and WL2. When you ran a batch of experiments in WL1 (e.g. via creating a new model version) it would submit each to the job queue before returning the next page, which tied up the server and made creation look like a very long job! Instead, WL2 creates the version immediately and stores in the DB a list of the experiments that should be submitted. This list is included in the JSON data about the new version, and the Javascript code then iterates through the list and submits each in turn in the background, adding notifications to the page as it goes. So much more responsive. But if they navigate away from the page before all are submitted, some will still be stuck in ‘pending’ state until the model or protocol is viewed again by someone with permission to run experiments.

Update suggested by (@jonc125 ) to remedy issue seen by @mirams :

> record the user who originally requested the experiment, and whenever they view an entity start submitting their experiments?

jonc125:
Fixes #244.


jonc125:
@sroderick-g5sro if you get a chance today (in a natural pause! No urgency) could you review this for me?

PR #249, opened 2020-02-04
Document how to update the local ontology copy

jonc125:
The last part of https://github.com/ModellingWebLab/deployment/pull/29


jonc125:
I've also added in here another aspect of avoiding open file errors. Deployment on scrambler now works again! Happy to merge @mirams ?

Issue #134, opened 2019-04-29
Add FittingResult model


jonc125:
Probably in a `fitting` app, cf #133?

See https://github.com/ModellingWebLab/project_issues/wiki/Workshop-notes-2018 for ideas.

Viewing results should hopefully just reuse code from #131 etc. But we'll need to do (at least some of) #135 before this can be fully implemented and tested.

The plan is being sketched out further in #203 and #241. Possible steps:

- [x] Do the `Runnable` refactor outlined in #241, with DB migrations, ensuring tests still pass etc. This can be a single PR.
    - This will mean that the `RunningExperiment` table won't need changing for fitting experiments, and indeed receipt of finished experiments in `experiments/processing.py:process_callback` (and cancelling running experiments) won't need any additions.
- [x] Then building on that and #250, create the `FittingResult` model inspired by the outline in #241. No views initially, just tests of the model directly? (Like `experiments/tests/test_models.py`)
    - So create the `FittingResult` and `FittingResultVersion` models.
    - `FittingResult` needs to link to repocache tables for entity versions, not use SHA strings.
- [x] Then basic views along the lines of `experiments/views.py`, trying to reuse code where possible as was done for the specs & datasets:
    - [x] `FittingResultVersionListView`
    - [x] `FittingResultVersionView`
    - [x] `FittingResultVersionJsonView`
    - [x] `FittingResultDeleteView`
    - [x] `FittingResultVersionDeleteView`
    - [x] `FittingResultFileDownloadView`
    - [x] `FittingResultVersionArchiveView`
- [x] Adapt `submit_experiment` to create `FittingResult` etc. instances, refactoring to share code where possible.
- [x] Harder views, probably need JS code rework e.g. as discussed in #131
    - [x] `FittingResultComparison[Json]View`

General notes:
* To get submission of fitting experiments to work we'll need a 'fitting' version of `experiments/processing.py:submit_experiment` that creates the `FittingResult` etc instances. We may be able to refactor so they share some code though.
* We don't want to try automatically migrating existing hacky fitting specs (stored as `Protocol` instances) to `FittingSpec` instances etc. Instead manually re-create ones we care about (just one really! i.e. Kylie since wave cell 5).
[...]

PR #253, opened 2020-02-27
Introduction of runnable base class to assist with introduction of fitting specs

sroderick-g5sro:
Part of #134


sroderick-g5sro:
Hi @jonc125. I am submitting this pull request for your review but I am aware of a number of issues:

1) I am not sure that I have taken the best course through the migrations and there maybe a number of redundant steps. 
2) In one step renaming ExperimentVersion to Runnable meant that I had to change all references in the code to generate the automatic migration. In particular I had to remove code in `experiments/admin.py` I have readded that and reverted the global reference changes. These steps didn't "feel" right and there may have been a simpler approach.
3) In migration ` 0028_auto_20200212_1701.py` I have added the logic to migrate the experiment_version records. I haven't added a reverse step to remove these - which I am guessing is required? Also I probably need to rename the file to something other than auto.
4) I have fixed the resulting unit test failures for example when fetching related experiments I have changed the code from `experiment_version__experiment` to `experiment_version__experimentversion__experiment`. I am assuming that I can do this double hop as the test pass. 
5) There is one test failure that I am still working on.  Experiment>test_sends_mail_when_experiment_is_finished

In general it would be good to get your feedback on the approach I have taken in this change.

weblab/experiments/apps.py
```diff
@@ -8,7 +8,7 @@ class ExperimentsConfig(AppConfig):
     name = 'experiments'
 
     def ready(self):
-        from .models import ExperimentVersion, RunningExperiment
+        from .models import Runnable, RunningExperiment
```
sroderick-g5sro: suspect this is wrong and I need to still reference ExperimentVersion here, but not sure why

weblab/experiments/forms.py
```diff
@@ -1,6 +1,6 @@
 from django import forms
 
-from .models import ExperimentVersion
+from .models import Runnable
```
sroderick-g5sro: .. and again here

weblab/experiments/migrations/0028_auto_20200212_1701.py
```diff
@@ -0,0 +1,27 @@
...
+        experiment_version.save()
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('experiments', '0027_experimentversion_experiment_key'),
+    ]
+
+    operations = [
+        migrations.RunPython(populate_exp_version),
```
sroderick-g5sro: not sure if I need a reverse migration as this is effectively a new table

weblab/experiments/migrations/0029_auto_20200212_1731.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.23 on 2020-02-12 17:31
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
```
jonc125: I think we could remove this migration.

weblab/experiments/migrations/0030_auto_20200212_1732.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.23 on 2020-02-12 17:32
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
```
jonc125: This one too.

weblab/experiments/migrations/0031_remove_runnable_experiment.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.23 on 2020-02-13 10:19
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
jonc125: This one is fine.

weblab/experiments/migrations/0032_experimentversion_experiment.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('experiments', '0031_remove_runnable_experiment'),
+    ]
+
+    operations = [
+        migrations.AddField(
```
jonc125: The models look like the right end state, but the migrations don't quite get us there reliably: I'm getting an error running them locally:
```
  Applying experiments.0032_experimentversion_experiment...Traceback (most recent call last):
  File "//anaconda3/envs/weblab/lib/python3.6/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
psycopg2.IntegrityError: insert or update on table "experiments_experimentversion" violates foreign key constraint "experiments_experime_experiment_id_5970cd0a_fk_experimen"
DETAIL:  Key (experiment_id)=(1) is not present in table "experiments_experiment".
```
Looks like it's the default value here that's causing issues, as I don't have an experiment with id 1 (it was deleted a while back...).

The issue I think is that you don't copy data from the `experiment_key` field to this new `experiment` field. If `experiment_key` was created as a proper `ForeignKey` initially, we could just have a `RenameField` migration here.

weblab/experiments/migrations/0033_remove_experimentversion_experiment_key.py
```diff
@@ -0,0 +1,20 @@
+
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.27 on 2020-02-27 12:12
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
jonc125: And if we renamed, we wouldn't need to delete, so this one would go.

weblab/experiments/migrations/0027_experimentversion_experiment_key.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.23 on 2020-02-12 17:00
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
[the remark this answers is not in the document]
jonc125: Yep, I think this would be the way forward. Rather than trying to create new automatic migrations, it might actually be easier to edit the current files by hand.

weblab/experiments/migrations/0027_experimentversion_experiment_key.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        ('experiments', '0026_experimentversion'),
+    ]
+
+    operations = [
+        migrations.AddField(
```
jonc125: So here you'd want
```
        migrations.AddField(
            model_name='experimentversion',
            name='experiment_key',
            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='experiments.Experiment', related_name='versions_copy'),
        ),
```

weblab/experiments/migrations/0028_auto_20200212_1701.py
```diff
@@ -0,0 +1,27 @@
...
+
+def populate_exp_version(apps, schema_editor):
+    ExperimentVersion = apps.get_model('experiments', 'ExperimentVersion')
+    Runnable = apps.get_model('experiments', 'Runnable')
+    for runnable in Runnable.objects.all():
+        experiment_version = ExperimentVersion(id=runnable.id,
+                                               created_at=runnable.created_at,
+                                               author_id=runnable.author_id,
+                                               experiment_id=runnable.experiment_id,
+                                               experiment_key=runnable.experiment_id)
```
jonc125: This would become `experiment_key_id=runnable.experiment_id`

weblab/experiments/migrations/0031_remove_runnable_experiment.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('experiments', '0030_auto_20200212_1732'),
```
jonc125: It would have to depend on 0028_auto_20200212_1701

weblab/experiments/migrations/0032_experimentversion_experiment.py
jonc125: So here we'd need
```
    operations = [
        migrations.RenameField(
            model_name='experimentversion',
            old_name='experiment_key',
            new_name='experiment',
        ),
        migrations.AlterField(
            model_name='experimentversion',
            name='experiment',
            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='versions', to='experiments.Experiment'),
        ),
    ]
```
```

File: weblab/experiments/migrations/0033_remove_experimentversion_experiment_key.py (hunk 8 above)
Comment (jonc125) [reply weblab/experiments/migrations/0033_remove_experimentversion_experiment_key.py]: For all the migrations to remove, we can just delete the file.

File: weblab/experiments/apps.py (hunk 1 above)
Comment (jonc125) [reply weblab/experiments/apps.py]: No, I think this one wants to be `Runnable` - it's making sure the results on disk are removed when the DB entry is deleted. And the results on disk are associated with `Runnable` now.

File: weblab/experiments/forms.py (hunk 2 above)
Comment (jonc125) [reply weblab/experiments/forms.py]: This one probably does want to be `ExperimentVersion`. It's for an admin-only interface that was created in the early dev days to aid debugging, before we had a back-end created, though. So we can possibly get rid of that functionality entirely! (But not in this PR...)

File: weblab/config/settings/base.py [hunk 13]
```diff
@@ -117,7 +117,7 @@
 
 # configure db url from DATABASE_URL env var if supplied
 DATABASES['default'] = dj_database_url.config(
-    default='postgres://weblab:weblab@localhost:5432/weblab'
+    default='postgres://postgres:postgres@localhost:5432/postgres5'
```
jonc125: I don't think you want to change this?

sroderick-g5sro: nope that was a mistake - thanks

PR #256, opened 2020-03-06
Annotation tool enhancements

### LINKED ISSUE #243: Tidy up output of metadata editor
A couple of things that would be nice.

1. At the moment WebLab metadata-annotated files look like this:

image

with all the RDF tags on one line. Whereas COR exports things that look like:
image
 which is much nicer with some formatting and indentation etc.

2. Also note how `gedit` picked up the highlighting for the COR exported file, but not ours, because the COR one has 
`<?xml version='1.0' encoding='UTF-8'?>` as the top line of the file, but we just start with the `<model>` tag.

### LINKED ISSUE #248: Allow all users to view (but not edit) the annotations of any model they can see
Now the annotations button is simply not shown. It'd be nicer if the users could still see if the model has the right annotations without having to look at the CellML code.

### LINKED ISSUE #255: Allow annotation with non-oxmeta terms
The logic to pull all terms used by protocols will reveal non-oxmeta terms that could be useful. The annotation tool should include a section containing these so that they can be added to models.

jonc125:
Fixes #255. Allow annotation with non-oxmeta terms
Fixes #243. Tidy up output of metadata editor
Fixes #248. Allow all users to view (but not edit) the annotations of any model they can see

### DIFF
--- a/weblab/static/js/visualizers/editMetadata/editMetadata.js
+++ b/weblab/static/js/visualizers/editMetadata/editMetadata.js
@@ -486,10 +486,15 @@ metadataEditor.prototype.saveNewVersion = function ()
     // Remove original RDF from model
     $(this.model.getElementsByTagNameNS("http://www.w3.org/1999/02/22-rdf-syntax-ns#", "RDF")).remove();
     // Dump updated RDF back into the model, and serialize (with base URI omitted)
+    // We make some modifications to the serialized XML so it's a bit tider
     var rdf_doc = this.modelRdf.databank.dump({format: 'application/rdf+xml', serialize: false});
     this.model.documentElement.appendChild(this.model.adoptNode(rdf_doc.documentElement));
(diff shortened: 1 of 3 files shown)


jonc125:
I've now deployed this on scrambler for you to test. Should allow:
- read-only view of annotations with the tool
- nicer output formatting when adding annotations to files
- annotating with non-oxmeta terms referenced by protocols

MichaelClerx:
> - read-only view of annotations with the tool
> - nicer output formatting when adding annotations to files
Checked these two and they look great!

> - annotating with non-oxmeta terms referenced by protocols
Don't get what's happened here. Where is the list of "custom" terms coming from? And how are they custom if they're already defined?

jonc125:
They're coming from the protocols (that you can see) in the system - any terms used in a model interface section. You can filter the list by protocol if you want to cut it down.

jonc125:
> we no longer get errors on conversion when models include non-oxmeta tags ?

Did we use to? I thought that aspect has been working for ages?!

PR #260, opened 2020-03-17
Template testing

jonc125:
@sroderick-g5sro I did a little playing while waiting for tests to run. Looks like the templates are rendered, but they don't check that reverse URL lookups actually exist! So you get the following in the HTML:

```
Corresponding model:\n        <a href="/entities/models/1/versions/mymodel1@9eecdce802a30253b89b3701c5f0d6006e7787a5">mymodel1 @ 9eecdce8...</a>
```

Instead of the expected:

```
Corresponding model:\n        <a href="/entities/models/1/versions/9eecdce802a30253b89b3701c5f0d6006e7787a5">mymodel1 @ 9eecdce8...</a>
```

Not sure how you'd check for that without having to hard-code every instance, but maybe you can find something!

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -16,8 +16,7 @@ defusedxml==0.5.0         # via python3-openid, social-auth-core
 dj-database-url==0.4.2
 django-braces==1.11.0
 django-guardian==1.4.9
-django==1.11.27
-first==2.0.2              # via pip-tools
+django==1.11.28
 flake8==3.4.1
 gitdb2==2.0.5             # via gitpython
 gitpython==2.1.7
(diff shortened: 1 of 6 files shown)


jonc125:
Hopefully this will make our tests more robust for the future!

sroderick-g5sro:
I think I see what you have done here ... set the defaults to have a space so it will fail if not set to a valid value? - if so happy to pass review. thanks for your help on this one.

jonc125:
Exactly, that's the idea. So models & protocols (etc) generated automatically by tests (except for the few tests that set a name specifically) will have a name that won't work in a URL, hopefully making tests fail wherever the deployed site would fail.

PR #285, opened 2020-09-16
changed reset password to link text

sroderick-g5sro:
Changed the reset password button to a text link - makes it less likely to be selected accidentally by users

### DIFF
--- a/weblab/templates/registration/login.html
+++ b/weblab/templates/registration/login.html
@@ -12,7 +12,7 @@
     {{ form.as_p }}
     <p>
       <button>Log in</button>
-      <a class="button" href="{% url "password_reset" %}">Reset Password</a>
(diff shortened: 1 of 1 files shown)


weblab/templates/registration/login.html
```diff
@@ -12,7 +12,7 @@ <h1>Log in</h1>
     {{ form.as_p }}
     <p>
       <button>Log in</button>
-      <a class="button" href="{% url "password_reset" %}">Reset Password</a>
+      <p><a href="{% url "password_reset" %}">Reset Password</a></p>
```
jonc125, line 15: Given we're already in a `<p>` here, maybe close that and then start a new one for the link? But otherwise looks fine!

jonc125 (approved):

Issue #272, opened 2020-08-01
Unfriendly errors can arise when creating initial entity version


jonc125:
See https://rollbar.com/jonc125/Web-Lab/items/86 - user sees error
> GitCommandError: Cmd('git') failed due to: exit code(128) cmdline: git reset --hard HEAD~ --
> stderr: 'fatal: Failed to resolve 'HEAD~' as a valid revision.'
On page https://scrambler.cs.ox.ac.uk/entities/models/442/versions/new

Probably because a git error happened while adding or removing a file, and the hard_reset then fails because no commit yet exists. Not sure what exactly the user did to cause this.

jonc125:
I figured out how to reproduce this while testing #286!

If you go through the process to create a new model, add a CellML file, enter an invalid tag name, and click 'create version' you'll get the error above.

Issue #229, opened 2020-01-13
Nicer error handling in new version of entity


mirams:
This is about when trying to make a new version of an entity, by deleting and upload a new protocol file.

If I do the file changes, n=but then do something that's not allowed (leave spaces in the tag name) then I get an error message (which could be a bit more helpful, it just says `Git command error: stderr: 'fatal: 'conductance can be dervied quantity' is not a valid tag name.'`, but all the changes I have done disappear. Would be nice to get the error message and keep the changes to the files.

jonc125:
I'm not sure how easy that would be to do, as it's a complex interaction between the JS code and Django file uploads. But possibly if we added some Django validation to the field to catch the most common errors, it would do something intelligent!

jonc125:
Relevant sections of code are:
* `entities/views.py:EntityNewVersionView`
* `static/js/upload.js`
* `templates/entities/includes/version_files.html` (and `upload.html` but I don't think that needs changing)

I suspect we might need to add logic in `get_context_data` that checks `request.POST` and adds `additions` and `deletions` to the context. And `version_files.html` can then use these to add extra rows for already-uploaded files, and add hidden `delete_filename[]` inputs for files marked for deletion.

However, we may wish to rethink how info is passed between the JS, form and view code, since it's already tricky to decipher cases where a user replaces an existing file, or uploads a file then decides it was a mistake and deletes it again before creating a new version, etc. So maybe having lists of `filenames[]` and `delete_filenames[]` isn't the best approach, and we should use a structure that reflects the order/detail of actions taken? Not sure yet what that could/should look like!

We also need to consider that changes impact dataset upload, since it uses the same javascript code (and equivalent template `templates/datasets/includes/dataset_files.html`).

PR #286, opened 2020-09-16
added validation on tagname

### LINKED ISSUE #261: Improve checking of tag names
Currently we delegate to the git command line for checking whether a tag name is allowed. We could do more checking in Django to give a friendly (& earlier) error. This would also let us give a much more tightly defined URL regex for entity versions. For instance, we could say tag names must be drawn from `[-_A-Za-z0-9]`. This is more constrained than [git allows](https://git-scm.com/docs/git-check-ref-format) but probably covers what users would actually want.

Then we could set tag form fields to validate against this regex too.

Follows from #260.

sroderick-g5sro:
Fixes #261.

### DIFF
--- a/weblab/entities/forms.py
+++ b/weblab/entities/forms.py
@@ -1,6 +1,7 @@
 from braces.forms import UserKwargModelFormMixin
 from django import forms
 from django.core.exceptions import ValidationError
+from django.core.validators import RegexValidator
 from django.forms import formset_factory
 
 from accounts.models import User
@@ -57,7 +58,7 @@ class EntityVersionForm(forms.Form):
     )
     tag = forms.CharField(
         help_text='Optional short label for this version',
-        required=False)
+        required=False, validators=[RegexValidator(r'[-_A-Za-z0-9]', 'Please enter a valid tag name.')])
     commit_message = forms.CharField(
         label='Description of this version',
         widget=forms.Textarea)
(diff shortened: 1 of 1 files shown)


weblab/entities/forms.py
```diff
@@ -57,7 +58,7 @@ class EntityVersionForm(forms.Form):
     )
     tag = forms.CharField(
         help_text='Optional short label for this version',
-        required=False)
+        required=False, validators=[RegexValidator(r'[-_A-Za-z0-9]', 'Please enter a valid tag name.')])
```
jonc125, line 61: Just a stylistic thing: it's best practice to have each arg on a separate line when they don't all fit on one.
```suggestion
        required=False,
        validators=[RegexValidator(r'[-_A-Za-z0-9]', 'Please enter a valid tag name.')])
```

weblab/entities/forms.py
```diff
@@ -94,7 +95,7 @@ class EntityTagVersionForm(forms.Form):
     tag = forms.CharField(
         label='New tag',
         help_text='Short label for this version',
-        required=True)
+        required=True, validators=[RegexValidator(r'[-_A-Za-z0-9]', 'Please enter a valid tag name.')])
```
jonc125, line 98:
```suggestion
        required=True,
        validators=[RegexValidator(r'[-_A-Za-z0-9]', 'Please enter a valid tag name.')])
```

jonc125 (changes requested):
I can confirm that this works as expected, albeit it doesn't solve #272 or #229 unfortunately. It would be good if the error message said what *was* valid, e.g. add `(only letters, numbers, dashes or underscores are allowed)`.

weblab/entities/forms.py
sroderick-g5sro: done

jonc125: The other occurrence is done, but you missed this one :)

sroderick-g5sro: sorry, changed that now

jonc125 (approved):
Great!

Issue #274, opened 2020-08-20
Add UI to link a dataset and protocol


jonc125:
This is to set up the mapping between dataset columns and protocol outputs (and inputs).

Initially this will be triggered on creation of a dataset, as the next step after adding files to it. But we could also use it to create a new link between already an existing protocol & dataset, e.g. once datasets are not tied to a single protocol. And we'll need it to be used for existing already-linked datasets to set up the mapping.

- [ ] Enhance the protocol analysis back-end task so it tells the front-end what the (inputs and) outputs are, and their units.
    - [x] Just outputs is enough for a first pass.
    - [x] On deployment of new back-end, re-analyse existing protocol versions.
    - [ ] Doing inputs will require adding info on their units, somewhere.
- [x] Cache that info in the DB, similarly to the interface a protocol expects from models.
- [x] Add a model for capturing the mapping from dataset columns to protocol inputs/outputs, probably with fields `dataset`, `column_name`, `column_units`, `protocol_version`, `protocol_in_out_put`, `protocol_units`, `is_output` or similar.
    - Actually, the last 3 can be replaced by a foreign key to `repocache.ProtocolIoputs`
- [x] Add a view to fill the mapping in, which gives essentially a row for each dataset column, with a text field allowing the units to be entered, and a drop-down selecting the protocol input/output (or not mapped)
    - [x] Need to consider how we want to allow the units to be specified. I suggest a text string that's parse-able by [pint](https://pint.readthedocs.io/).
    - [x] Jonathan to give examples of pint definition strings and the parsing process.

jonc125:
Quickstart tips for [pint unit processing](https://pint.readthedocs.io/en/stable/tutorial.html#tutorial):

```python
from pint import UnitRegistry
ureg = UnitRegistry()

quantity = ureg(unit_string)  # Can raise pint.UndefinedUnitError, pint.DefinitionSyntaxError - should give a friendly form error for each type separately I think

display_for_templates = '{:H}'.format(quantity)
```

jonc125:
[...]

PR #287, opened 2020-09-17
Store analysis of protocol inputs & outputs

jonc125:
This solves the backend analysis part of #274. Note that the backend only ever returns protocol outputs at present, but the frontend should be capable of receiving inputs once that is added.

Testing requires deploying the corresponding backend branches using https://github.com/ModellingWebLab/deployment/tree/run-weblab-fc.

cf https://github.com/ModellingWebLab/fc-runner/pull/10 and https://github.com/ModellingWebLab/weblab-fc/pull/214.


helenst (approved):
This looks good to me! :+1:

helenst (approved):

helenst:
Looks good to me!

Issue #247, opened 2020-02-03
No sharing (collaborators) functionality for datasets


sroderick-g5sro:
Other entities (including fitting specs) have a manage collaborators functionality that allows sharing. This has not been implemented for datasets. 

Background:
from a coding perspective, datasets should inherit the add_collaborators method but the guardian library throws a permissions issue, eg, 

```
        if not num:
            raise self.model.DoesNotExist(
                "%s matching query does not exist." %
>               self.model._meta.object_name
            )
E           django.contrib.auth.models.DoesNotExist: Permission matching query does not exist.
```

a new version of this library exists but required Django version 2

jonc125:
I think I've figured out how to do the permissions properly - see my 2 commits on the PR. Which means (apart from fixing the tests that the change to `is_editable_by` broke!) that the remaining steps are the UI for adding dataset collaborators.

It should be possible to use `EntityCollaboratorForm` and `EntityCollaboratorFormSet` directly, so no need to write a new form. `datasets.views.py` will need a view inheriting from `EntityCollaboratorsView` that changes the `template_name` field and `get_context_data` (to remove the `type` kwarg). If the new view is accessed via a URL named `datasets:entity_collaborators` then nothing else should need changing in the code. Just a new template added based on the entity collaborator one but adapted for datasets.

I'd suggest a bit of experimentation with the feature live to see if any oddities arise (check a test user you share with can see the dataset, etc.) and review the collaboration tests for entities to see which ones make sense to replicate for datasets.

sroderick-g5sro:
When I click on the Manage Collaborators link I get the following error message. I dont understand why because it refers to a link in the dataset header.

![image](https://user-images.githubusercontent.com/3098326/100448616-77f43a00-30aa-11eb-9304-bdbeee81d7b0.png)

jonc125:
Looks like that's because `dataset` isn't provided in the context for the template, and the header expects it. So `DatasetCollaboratorsView` probably needs its own `get_context_data` that sets that as well as calling the superclass method.

sroderick-g5sro:
thanks Jonathan, I'm not clear now I set the dataset and call the superclass method. I think you mention before not using kwargs so this is as far as I got

```
    def get_context_data(self):
        dataset = self.get_object()
        return super().get_context_data()
```

jonc125:
Follow an analogous approach to `EntityCollaboratorsView` - using `kwargs` is appropriate here as that is the interface Django expects the method to have.

PR #288, opened 2020-09-17
167 allow renaming

### LINKED ISSUE #167: Allow renaming of entities
It's quite common to think of a better name for a model / protocol further down the line

sroderick-g5sro:
Will now fix #167.


weblab/entities/forms.py
```diff
@@ -97,6 +97,21 @@ class EntityTagVersionForm(forms.Form):
...
+        label='New Name',
+        help_text='',
+        required=True)
+
+    def clean_name(self):
+        name = self.cleaned_data['name']
+        if Entity.objects.filter(name=name).exists():
+            raise ValidationError(
+                'You already have a model or protocol named "%s"' % name)
+
```
sroderick-g5sro, line 111: Ideally would say if it was a model or protocol but unsure how to get the display type here

weblab/entities/views.py
```diff
@@ -778,6 +779,42 @@ def get(self, request, *args, **kwargs):
...
+    context_object_name = 'entity'
+    form_class = EntityRenameForm
+
+    def _get_object(self):
+        if not hasattr(self, 'object'):
+            self.object = self.get_object()
+        return self.object
+
+    def test_func(self):
+        return self._get_object().is_managed_by(self.request.user)
```
jonc125, line 793: `is_editable_by` would probably be better than `is_managed_by` here, assuming we want to allow collaborators with write access to rename entities @mirams ?

weblab/entities/views.py
```diff
@@ -778,6 +779,42 @@ def get(self, request, *args, **kwargs):
         return response
 
 
+class RenameView(LoginRequiredMixin, UserPassesTestMixin, FormMixin, EntityTypeMixin, DetailView):
+    template_name = 'entities/entity_rename_form.html'
+    context_object_name = 'entity'
+    form_class = EntityRenameForm
```
jonc125, line 785: Similarly to `EntityCreateView`, you can have 2 subclasses of `EntityRenameForm`, one for models and one for protocols (and a third for `FittingSpec` in the `fitting` app actually). And rather than defining `form_class` like this, make it an `@property` which returns the form class corresponding to the class `self.model`.

weblab/entities/views.py
jonc125: Which means you could just inherit from `EntityCollaboratorRequiredMixin` instead of `UserPassesTestMixin` and remove this method.

weblab/entities/views.py
```diff
@@ -778,6 +779,42 @@ def get(self, request, *args, **kwargs):
...
+
+    def _get_object(self):
```
jonc125, line 787: The `DetailView` class sets `self.object` so we don't need this method.

weblab/entities/views.py
```diff
@@ -778,6 +779,42 @@ def get(self, request, *args, **kwargs):
...
+    def post(self, request, *args, **kwargs):
+        """Check the form and possibly add the tag in the repo.
+
+        Called by Django when a form is submitted.
+        """
+        form = self.get_form()
+
+        if form.is_valid():
+            new_name = form.cleaned_data['name']
+            entity = self._get_object()
```
jonc125, line 804:
```suggestion
            entity = self.object
```

weblab/entities/forms.py
jonc125: Indeed, it's rather awkward! The trick is used by `EntityCreateView` (see also below) and the two subclasses of `EntityForm` - you basically need a form subclass for each entity type.

jonc125 (changes requested):
I can confirm that this works nicely on my system.

I've added a few comments, particularly around how to make the name uniqueness check specific to the entity type. Since it's fine to own a model & protocol with the same name, just not 2 models or 2 protocols.

The other thing needed before this can be merged is a few automated tests: for each of a model and a protocol, check both successful rename and error message case.

Then probably merge this PR, and do renaming of fitting specs & datasets in a separate PR to close the issue? Or those features can be added to this PR if you prefer.

weblab/entities/views.py
sroderick-g5sro: When I try and take this out I get an error message saying the method is required

jonc125: That might be just because it's used in the `post` and `test_func` methods below? In which case implementing the other changes would remove the need for it. Or it might be that one of the base classes needs it, in which case leave it in.

weblab/entities/forms.py
```diff
@@ -6,7 +6,12 @@
 from accounts.models import User
 from core import visibility
 
-from .models import EntityFile, ModelEntity, ProtocolEntity
+from .models import (
+    Entity,
```
jonc125, line 10: You can get rid of this import now. There are a few other flake8 errors in the build too; I'll not go through them all.

weblab/entities/forms.py
```diff
@@ -31,6 +36,29 @@ def save(self, **kwargs):
     def entity_type(self):
         return self._meta.model.entity_type
 
+class EntityRenameForm(UserKwargModelFormMixin, forms.ModelForm):
+    """Used for renaming an existing entity."""
+
+    def clean_name(self):
+        name = self.cleaned_data['name']
+        if self._meta.model.objects.filter(name=name).exists():
```
jonc125, line 44: I thought we might need a `, author=self.user` argument in the filter too, but `EntityForm` doesn't have this. So I think there may be a missing test case for creation too, that checks it's OK for 2 users to have a model with the same name!

weblab/entities/tests/test_views.py
```diff
@@ -84,11 +84,75 @@ def test_create_protocol_requires_permissions(self, logged_in_user, client):
...
+                 })
+        assert response.status_code == 302
+        entity = ModelEntity.objects.first()
+        assert entity.name == 'new name'
+
+    def test_model_renaming_failure(self, client, logged_in_user, helpers):
+        helpers.add_permission(logged_in_user, 'create_model')
+        model = recipes.model.make(author=logged_in_user)
+
+        model2 = recipes.model.make(name='test model 2')
```
jonc125, line 107: Yeah, following from the comment above: this shouldn't fail, because `model2` has a different author from `model`.

weblab/entities/tests/test_views.py
```diff
@@ -84,11 +84,75 @@ def test_create_protocol_requires_permissions(self, logged_in_user, client):
...
+        commit = helpers.add_version(model, visibility='private')
+
+        abs_path = model.repo_abs_path
+        response = client.post(
+            '/entities/models/%d/rename' % model.pk,
+            data={
+                     'name': 'test protocol'
+                 })
+        assert response.status_code == 302
+        assert abs_path == model.repo_abs_path
```
jonc125, line 148: `model` here won't actually reflect changes in the database. So instead you want to check `entity.repo_abs_path` after checking the name.

weblab/entities/tests/test_views.py
```diff
@@ -84,11 +84,75 @@ def test_create_protocol_requires_permissions(self, logged_in_user, client):
...
+        entity = ModelEntity.objects.first()
+        assert entity.name == 'test protocol'
 
 @pytest.mark.django_db
 @pytest.mark.parametrize("recipe,url,list_url", [
     (recipes.model, '/entities/models/%d/delete', '/entities/models/'),
-    (recipes.protocol, '/entities/protocols/%d/delete', '/entities/protocols/'),
+    (recipes.protocol, '/entities/protocols/%d/delete', '/entit`1ies/protocols/'),
```
jonc125, line 155: Oops!

weblab/entities/views.py
```diff
@@ -936,8 +979,8 @@ def _get_bives_diff(self, file1, file2):
                 result['response'] = True
         else:
             result['responseText'] = (
-                'bives request failed: %d (%s)' %
-                (bives_response.status_code, bives_response.content.decode())
+                    'bives request failed: %d (%s)' %
+                    (bives_response.status_code, bives_response.content.decode())
```
jonc125, line 983:
```suggestion
                'bives request failed: %d (%s)' %
                (bives_response.status_code, bives_response.content.decode())
```

weblab/entities/forms.py
sroderick-g5sro: sounds like we do need to filter by username in the form so I have added `author=self.user` as a filter option - is this correct?

sroderick-g5sro: interestingly I now get a database integrity error thrown, when I try to a rename model to one that exists. Not sure if I need to trap this, was just expecting the message from the form
 
image

weblab/entities/tests/test_views.py
sroderick-g5sro: thanks, change made

weblab/entities/forms.py
jonc125: Interesting. `Entity` definitely has `unique_together = ('entity_type', 'name', 'author')` so should allow same name different author. And it looks like from the error it's complaining that the author also matches. So as you say I'd expect the form validation to catch this before it hits the DB. Could you push/paste your work-in-progress so I can take a look?

jonc125: The test looks OK, so it might be something in this file that's wrong, but I don't think you've pushed any changes here?

weblab/entities/forms.py
```diff
@@ -32,6 +37,30 @@ def entity_type(self):
         return self._meta.model.entity_type
 
 
+class EntityRenameForm(UserKwargModelFormMixin, forms.ModelForm):
+    """Used for renaming an existing entity."""
+
+    def clean_name(self):
+        name = self.cleaned_data['name']
+        if self._meta.model.objects.filter(author=self.user, name=name).exists():
```
jonc125, line 45: This looks like what I'd expect... so not sure why it isn't working! Maybe put some print calls in to see whether it's actually being used, and what the DB contains here if it is?

sroderick-g5sro: thanks Jonathan, it appears to run this line fine.

sroderick-g5sro: in the database are the two models my model1, test model 2 so still not clear why getting the database constraint issue

jonc125: It looks like `self.user` is `None` in the form, so the filter always returns false. The patch below resolves the issue; I can't commit it locally because of flake8 issues elsewhere! Basically you need to add `UserFormKwargsMixin` as a superclass for the view so it passes the user through to the form.

```diff
diff --git c/weblab/entities/tests/test_views.py i/weblab/entities/tests/test_views.py
index b1d0c81..31042b4 100644
--- c/weblab/entities/tests/test_views.py
+++ i/weblab/entities/tests/test_views.py
@@ -146,9 +146,9 @@ class TestEntityRenaming:
             data={
                      'name': 'test model 2'
                  })
-        assert response.status_code == 302
+        assert response.status_code == 200
         entity = ModelEntity.objects.first()
-        assert entity.name == 'test model 2'
+        assert entity.name == 'my model1'

     def test_model_and_protocol_renaming_success(self, client, logged_in_user, helpers):
         helpers.add_permission(logged_in_user, 'create_model')
@@ -186,7 +186,7 @@ class TestEntityRenaming:
 @pytest.mark.django_db
 @pytest.mark.parametrize("recipe,url,list_url", [
     (recipes.model, '/entities/models/%d/delete', '/entities/models/'),
-    (recipes.protocol, '/entities/protocols/%d/delete', '/entit1ies/protocols/'),
+    (recipes.protocol, '/entities/protocols/%d/delete', '/entities/protocols/'),
 ])
 class TestEntityDeletion:
     def test_owner_can_delete_entity(
diff --git c/weblab/entities/views.py i/weblab/entities/views.py
index d93fa89..c87005c 100644
--- c/weblab/entities/views.py
+++ i/weblab/entities/views.py
@@ -779,7 +779,7 @@ class EntityFileDownloadView(EntityTypeMixin, EntityVersionMixin, SingleObjectMi
         return response


-class RenameView(LoginRequiredMixin, UserPassesTestMixin, FormMixin, EntityTypeMixin, DetailView):
+class RenameView(LoginRequiredMixin, UserPassesTestMixin, EntityTypeMixin, UserFormKwargsMixin, FormMixin, DetailView):
     template_name = 'entities/entity_rename_form.html'
     context_object_name = 'entity'

```

weblab/datasets/forms.py
```diff
@@ -32,6 +32,24 @@ def save(self, **kwargs):
         self.save_m2m()
         return dataset
 
+class EntityRenameForm(UserKwargModelFormMixin, forms.ModelForm):
```
jonc125, line 35: There's no need to have subclasses here; this can just be `DatasetRenameForm`

weblab/datasets/views.py
```diff
@@ -230,3 +230,41 @@ def test_func(self):
...
+        if self.model is Dataset:
+            return DatasetRenameForm
+
+    def _get_object(self):
+        if not hasattr(self, 'object'):
+            self.object = self.get_object()
+        return self.object
+
+    def test_func(self):
+        return self._get_object().is_deletable_by(self.request.user)
```
jonc125, line 254: `is_managed_by` might be better, although it's the same effect! `is_editable_by` doesn't yet work for datasets due to a quirk in Django's permission system, because they're not in the entities app. See also #247.

weblab/fitting/forms.py
```diff
@@ -24,3 +26,20 @@ class FittingSpecVersionForm(EntityVersionForm):
     This works almost the same as other entities, except we can't re-run experiments.
     """
     rerun_expts = None
+
+class EntityRenameForm(UserKwargModelFormMixin, forms.ModelForm):
```
jonc125, line 30: You don't want to repeat this class here, but import it from `entities.forms`.

weblab/fitting/urls.py
```diff
@@ -132,4 +132,10 @@
         entity_views.EntityDiffView.as_view(),
         name='diff',
     ),
+    url(
+        r'^%s/(?P<pk>\d+)/rename$' % _ENTITY_TYPE,
+        views.RenameView.as_view(),
```
jonc125, line 137: I think this should be:
```suggestion
        entity_views.RenameView.as_view(),
```

jonc125: Actually no, because you need a different form. So you'll need to create a `FittingSpecRenameView` class in `fitting/views.py` that inherits from `entities.views.RenameView`, and reference `views.FittingSpecRenameView` here.

weblab/datasets/tests/test_views.py
```diff
@@ -113,6 +113,55 @@ def test_add_delete_add_file(self, logged_in_user, client, helpers, public_proto
                 assert fp.read() == file_map[f.name]
 
 
+@pytest.mark.django_db
+class TestEntityRenaming:
```
jonc125, line 117:
```suggestion
class TestDatasetRenaming:
```

weblab/datasets/views.py
```diff
@@ -230,3 +235,37 @@ def test_func(self):
...
+
+    def test_func(self):
+        return self._get_object().is_managed_by(self.request.user)
+
+    def post(self, request, *args, **kwargs):
+        form = self.get_form()
+
+        if form.is_valid():
+            new_name = form.cleaned_data['name']
+            dataset = self.get_object()
```
jonc125, line 262:
```suggestion
            dataset = self._get_object()
```

weblab/datasets/views.py
```diff
@@ -230,3 +235,37 @@ def test_func(self):
...
+            dataset.name = new_name
+            dataset.save()
+            return self.form_valid(form)
+        else:
+            return self.form_invalid(form)
+
+    def get_success_url(self, *args, **kwargs):
+        ns = self.request.resolver_match.namespace
+        return reverse(ns + ':list')
```
jonc125, line 271: Probably better for this to redirect to the renamed dataset. So
```suggestion
        return reverse(ns + ':detail', args=[self._get_object().id])
```
I think.

weblab/entities/tests/test_views.py
```diff
@@ -51,6 +51,23 @@ def test_create_model(self, logged_in_user, client, helpers):
 
         assert entity.repo_abs_path.exists()
 
+    def test_create_model_same_name(self, logged_in_user, other_user, client, helpers):
```
jonc125, line 54: I don't think this is checking anything new at present, as it doesn't create two models? We probably want to defer the fixing of the bug on entity creation for another PR, so I'd suggest removing this test for now.

weblab/datasets/tests/test_views.py
```diff
@@ -113,6 +113,55 @@ def test_add_delete_add_file(self, logged_in_user, client, helpers, public_proto
...
+
+        response = client.post(
+            '/datasets/%d/rename' % dataset.pk,
+            data={
+                'name': 'new name'
+            })
+        assert response.status_code == 302
+        dataset = Dataset.objects.first()
+        assert dataset.name == 'new name'
+
```
jonc125, line 131: We should check that after the rename the dataset's archive still can be found on disk, since it behaves differently from models & protocols. So here we want
```python
        assert dataset.archive_path.exists()
```
Which means you need a dataset with a file in the first place. So rather than creating a dataset in the test I suggest using the `my_dataset_with_file` fixture (and so no need for `logged_in_user` or `helpers`). And assert that `dataset.archive_path.exists()` before the rename too just in case!

weblab/datasets/views.py
```diff
@@ -230,3 +235,37 @@ def test_func(self):
...
+    def test_func(self):
+        return self._get_object().is_managed_by(self.request.user)
+
+    def post(self, request, *args, **kwargs):
+        form = self.get_form()
+
+        if form.is_valid():
+            new_name = form.cleaned_data['name']
+            dataset = self.get_object()
+            dataset.name = new_name
```
jonc125, line 263: You'll probably need to implement the archive name here. Something like:
```suggestion
            old_archive_path = dataset.archive_path
            dataset.name = new_name
            if old_archive_path.exists():
                old_archive_path.rename(dataset.archive_path)
```

weblab/entities/views.py
```diff
@@ -778,6 +780,48 @@ def get(self, request, *args, **kwargs):
...
+    def _get_object(self):
+        if not hasattr(self, 'object'):
+            self.object = self.get_object()
+        return self.object
+
+    def test_func(self):
+        return self._get_object().is_editable_by(self.request.user)
+
+    def post(self, request, *args, **kwargs):
+        """Check the form and possibly add the tag in the repo.
```
jonc125, line 803:
```suggestion
        """Check the form and possibly rename the entity.
```

jonc125 (changes requested):
Looking good! There are a couple of things that could do with checking before merging though.

weblab/datasets/tests/test_views.py
sroderick-g5sro: done - thanks

weblab/datasets/tests/test_views.py
```diff
@@ -114,19 +114,22 @@ def test_add_delete_add_file(self, logged_in_user, client, helpers, public_proto
...
 @pytest.mark.django_db
-class TestEntityRenaming:
-    def test_dataset_renaming_success(self, client, logged_in_user, helpers):
+class TestDatasetRenaming:
+    def test_dataset_renaming_success(self, client, my_dataset_with_file):
 
-        dataset = recipes.dataset.make(author=logged_in_user)
-        assert dataset.name == 'my dataset1'
+        dataset = my_dataset_with_file
+        assert dataset.archive_path.exists()
```
jonc125, line 121: We could make this test slightly stronger with this change and the next below.
```suggestion
        old_path = dataset.archive_path
        assert old_path.exists()
```

weblab/datasets/tests/test_views.py
```diff
@@ -114,19 +114,22 @@ def test_add_delete_add_file(self, logged_in_user, client, helpers, public_proto
...
 
         response = client.post(
             '/datasets/%d/rename' % dataset.pk,
             data={
                 'name': 'new name'
             })
         assert response.status_code == 302
+
         dataset = Dataset.objects.first()
+        assert dataset.archive_path.exists()
```
jonc125, line 132:
```suggestion
        assert not old_path.exists()
        assert dataset.archive_path.exists()
```

jonc125 (changes requested):
This looks good in terms of functionality. We should just add tests for the fitting spec case, in `fitting/tests/test_views.py`, along the same lines as those added to `entities/tests/test_views.py`.

sroderick-g5sro:
> This looks good in terms of functionality. We should just add tests for the fitting spec case, in `fitting/tests/test_views.py`, along the same lines as those added to `entities/tests/test_views.py`.

weblab/fitting/tests/test_views.py
```diff
@@ -0,0 +1,53 @@
+import pytest
+
+from weblab.core import recipes
```
jonc125, line 3:
```suggestion
from core import recipes
```
is used elsewhere

jonc125 (approved):
Looks ready to merge!

Issue #277, opened 2020-08-27
UI to run a new fitting experiment


jonc125:
Recall that a fitting spec is linked to a protocol (later maybe more than one) which is in turn linked to zero or more datasets.
To be able to run a fit we need to specify:
* Version of fitting spec
* Version of protocol
* Dataset
* Version of model

This view is reached initially via a button from viewing a single version of a fitting spec.
Later, via a button from a dataset (iff there is a fitting spec connected to a linked protocol) or model.

The view should have 4 drop-down menus in a row, to select Fitting spec, Protocol, Dataset, Model.
One of these will be pre-selected and unable to be changed, depending on how the view was reached.

Selecting options from each drop-down will constrain what options are offered in the others, with Javascript code updating the lists.
* Once a fitting spec is chosen, only linked protocols and datasets are available. Similarly if we start from a dataset.
* Initially all models are listed, but eventually these will be filtered to show only those with an interface compatible to the selected protocol & fitting spec (once this interface info is in the DB)

Under each of the top drop-downs with versions (i.e. not datasets) there should be a drop-down to select version, defaulting to the latest.
This should be greyed out until a selection is made in the upper row, and then filled in via an AJAX call with the available versions.

At the bottom there should be a button 'Fit model to data' which submits the fitting experiment to the queue.

The submission & result retrieval system will be very similar to that for normal experiments; common code should be extracted. See also #134.
See the `if protocol.is_fitting_spec` branch in `experiments/processing.py` for what needs to be sent to the back-end for a fitting experiment.
It might make sense to change `submit_experiment` so it takes `model_version` etc as a `CachedModelVersion`, drops the `model` & `protocol` arguments (since these can be found from the cached versions) and has optional `dataset` and `fitting_spec_version` arguments?

Child of #135.

Tasks:
- [x] Create basic view with form
- [x] Back end validation of valid entity / version combinations
- [x] Allow preselection of entities and versions on page load
- [x] Check user permissions on entities (preselected and submitted)
- [x] Check user permissions on ajax endpoint
- [x] Submit fitting experiment to back end
- [x] Simplify call signature for submission code
- [x] Javascript filtering of dropdown values
  - [x] Grey out versions until entity selected
  - [x] Ajax endpoint to provide valid values
- [x] Page and form styling
- [x] Buttons from entity pages
- [x] The JS action to re-run a fitting experiment doesn't work - it tries to do a normal experiment run, not a fitting experiment

mirams:
[...]

PR #290, opened 2020-09-22
UI to run fitting experiments

helenst:
Addresses #277

### DIFF
--- a/weblab/fitting/tests/test_views.py
+++ b/weblab/fitting/tests/test_views.py
@@ -475,6 +475,22 @@ class TestCreateFittingResultView:
         assert response.status_code == 200
         assert response.context['form'].initial['dataset'] == public_dataset
 
+    def test_with_non_visible_model(self, client, fits_user, private_model):
+        response = client.get('/fitting/results/new', {'model': private_model.pk})
+        assert response.status_code == 404
+
+    def test_with_non_visible_protocol(self, client, fits_user, private_protocol):
+        response = client.get('/fitting/results/new', {'protocol': private_protocol.pk})
+        assert response.status_code == 404
+
+    def test_with_non_visible_fittingspec(self, client, fits_user, private_fittingspec):
+        response = client.get('/fitting/results/new', {'fittingspec': private_fittingspec.pk})
+        assert response.status_code == 404
+
+    def test_with_non_visible_dataset(self, client, fits_user, private_dataset):
+        response = client.get('/fitting/results/new', {'dataset': private_dataset.pk})
+        assert response.status_code == 404
+
     @patch('fitting.views.submit_fitting')
     def test_submits_to_backend(self, mock_submit, client, fits_user, public_model, public_protocol,
                                 public_fittingspec, public_dataset, helpers):
--- a/weblab/fitting/tests/test_views.py
+++ b/weblab/fitting/tests/test_views.py
@@ -680,7 +680,7 @@ class TestFittingResultFilterJsonView:
         options = data['fittingResultOptions']
         assert set(options['datasets']) == {dataset1.id}
 
-    def test_model_and_versions_restricted_when_model_selected(self, client, fits_user, helpers):
+    def test_versions_restricted_when_model_selected(self, client, fits_user, helpers):
         model1 = recipes.model.make()
         model2 = recipes.model.make()
         m1v1 = helpers.add_cached_version(model1, visibility='public')
@@ -691,10 +691,10 @@ class TestFittingResultFilterJsonView:
 
         data = json.loads(response.content.decode())
         options = data['fittingResultOptions']
-        assert options['models'] == [model1.id]
+        assert options['models'] == [model1.id, model2.id]
         assert options['model_versions'] == [m1v1.id]
 
-    def test_protocol_and_versions_restricted_when_protocol_selected(self, client, fits_user, helpers):
+    def test_versions_restricted_when_protocol_selected(self, client, fits_user, helpers):
         protocol1 = recipes.protocol.make()
         protocol2 = recipes.protocol.make()
         p1v1 = helpers.add_cached_version(protocol1, visibility='public')
@@ -705,10 +705,10 @@ class TestFittingResultFilterJsonView:
 
         data = json.loads(response.content.decode())
         options = data['fittingResultOptions']
-        assert options['protocols'] == [protocol1.id]
+        assert options['protocols'] == [protocol1.id, protocol2.id]
         assert options['protocol_versions'] == [p1v1.id]
 
-    def test_fittingspec_and_versions_restricted_when_fittingspec_selected(self, client, fits_user, helpers):
+    def test_versions_restricted_when_fittingspec_selected(self, client, fits_user, helpers):
         fittingspec1 = recipes.fittingspec.make()
         fittingspec2 = recipes.fittingspec.make()
         f1v1 = helpers.add_cached_version(fittingspec1, visibility='public')
@@ -719,20 +719,8 @@ class TestFittingResultFilterJsonView:
 
         data = json.loads(response.content.decode())
         options = data['fittingResultOptions']
-        assert options['fittingspecs'] == [fittingspec1.id]
+        assert options['fittingspecs'] == [fittingspec1.id, fittingspec2.id]
         assert options['fittingspec_versions'] == [f1v1.id]
-
-    def test_dataset_restricted_when_selected(self, client, fits_user, helpers):
-        dataset1 = recipes.dataset.make(visibility='public')
-        dataset2 = recipes.dataset.make(visibility='public')  # noqa: F841
-
-        response = client.get('/fitting/results/new/filter', {'dataset': dataset1.id})
-        assert response.status_code == 200
-
-        data = json.loads(response.content.decode())
-        assert 'fittingResultOptions' in data
-        options = data['fittingResultOptions']
-        assert options['datasets'] == [dataset1.id]
 
     def test_dataset_and_fittingspec_restricted_when_protocol_selected(self, client, fits_user, helpers):
         protocol = recipes.protocol.make()
--- a/weblab/fitting/tests/test_views.py
+++ b/weblab/fitting/tests/test_views.py
@@ -453,3 +453,23 @@ class TestCreateFittingResultView:
         response = client.get('/fitting/results/new')
         assert response.status_code == 200
         assert 'form' in response.context
+
+    def test_with_preselected_model(self, client, fits_user, public_model):
+        response = client.get('/fitting/results/new', {'model': public_model.pk})
+        assert response.status_code == 200
+        assert response.context['form'].initial['model'] == public_model
+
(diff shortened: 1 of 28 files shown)


weblab/fitting/views.py
```diff
@@ -197,3 +204,150 @@ def get(self, request, *args, **kwargs):
...
+    """
+    Create and submit a fitting result from models, protocols, fitting specs and datasets
+    and (where relevant) their versions.
+    """
+    permission_required = 'fitting.run_fits'
+    form_class = FittingResultCreateForm
+
+    template_name = 'fitting/fittingresult_create_form.html'
+
+    def get_initial(self):
```
jonc125, line 219: Should we check visibility here, in case e.g. one user sends a URL to another that includes IDs the receiver isn't able to see?

weblab/fitting/views.py
```diff
@@ -197,3 +204,150 @@ def get(self, request, *args, **kwargs):
...
+        elif 'protocol' in self.request.GET:
+            initial['protocol'] = get_object_or_404(ProtocolEntity, pk=self.request.GET['protocol'])
+        elif 'fittingspec' in self.request.GET:
+            initial['fittingspec'] = get_object_or_404(FittingSpec, pk=self.request.GET['fittingspec'])
+        elif 'dataset' in self.request.GET:
+            initial['dataset'] = get_object_or_404(Dataset, pk=self.request.GET['dataset'])
+
+        return initial
+
+    def post(self, request, *args, **kwargs):
```
jonc125, line 232: Does the base class not provide this implementation as default? I think in other views we've done the logic in `post` rather than implementing `form_valid`; not sure what is considered Django best practice here?

weblab/fitting/views.py
```diff
@@ -197,3 +204,150 @@ def get(self, request, *args, **kwargs):
...
+                messages.info(self.request, "Fitting experiment submitted to the queue.")
+            else:
+                messages.error(self.request, "Fitting experiment could not be run: " + self.runnable.return_text)
+        else:
+            messages.info(self.request, "Fitting experiment was already run.")
+
+        return super().form_valid(form)
+
+    def get_success_url(self):
+        if hasattr(self, 'runnable'):
```
jonc125, line 262: Is it possible for `self.runnable` not to exist? And what happens if so?

weblab/fitting/views.py
```diff
@@ -197,3 +204,150 @@ def get(self, request, *args, **kwargs):
...
+
+    Connections between protocols, fitting specs and datasets are also enforced.
+    """
+    permission_required = 'fitting.run_fits'
+
+    def get(self, request, *args, **kwargs):
+        options = {}
+
+        models = ModelEntity.objects.visible_to_user(request.user)
+        model_versions = CachedModelVersion.objects.visible_to_user(request.user)
```
jonc125, line 283: Given we're doing the visibility query to get versions, would it be more efficient to make `models` a query based on `model_versions` etc? Something like `models = ModelEntity.objects.filter(cachedmodel__versions__in=model_versions)`?

weblab/fitting/views.py
```diff
@@ -197,3 +204,150 @@ def get(self, request, *args, **kwargs):
...
+
+        # Restrict to specified model and its versions
+        if model:
+            models = [model]
+            model_versions = model_versions.filter(entity__entity=model)
+
+        # Restrict to specified fitting spec and its versions
+        if fittingspec:
+            fittingspecs = [fittingspec]
+            fittingspec_versions = fittingspec_versions.filter(entity__entity=fittingspec.id)
```
jonc125, line 318: Is there a reason `.id` is used here and not for model above?

weblab/fitting/views.py
```diff
@@ -197,3 +204,150 @@ def get(self, request, *args, **kwargs):
...
+        options['models'] = _get_ids(models)
+        options['model_versions'] = _get_ids(model_versions)
+        options['protocols'] = _get_ids(protocols)
+        options['protocol_versions'] = _get_ids(protocol_versions)
+        options['fittingspecs'] = _get_ids(fittingspecs)
+        options['fittingspec_versions'] = _get_ids(fittingspec_versions)
+        options['datasets'] = _get_ids(datasets)
+
+        return JsonResponse({
+            'fittingResultOptions': options
```
jonc125, line 352: I'm not sure where this info is fed to the user-visible page, but it'd be nice if the drop-downs for entity version displayed a more user-friendly identifier than e.g. "my model@very long sha"! The entity name is redundant since the user has selected an entity just above. And tag names would be preferable to SHAs where they exist, as is done when viewing an entity version.

weblab/templates/entities/entity_version.html
```diff
@@ -63,10 +63,14 @@ <h2 id="entityversionname">
         {% endif %}
         <li><a href="{% entity_url 'version_list' entity %}" title="Show differences between versions of this {{ type }}">Other versions</a></li>
       </ul></div>
-      <!--
-      {% if type == 'model' %}
-        <a class="button" href="" title="Fit the parameters of this model to data">Fit</a>
+      {% if entity.entity_type == 'model' %}
+        <a class="button" href="{% url "fitting:result:new" %}?model={{entity.pk}}" title="Fit the parameters of this model to data">Fit</a>
```
jonc125, line 67: It would be nice to include the model version (id or sha) as an extra query parameter here (and similar for protocol & fitting spec below), so that it's filled in as the default in the form. Maybe allow the user still to change the version, but not the entity itself?

jonc125 (commented):
This is looking good! I've played around with it live and set 4 fitting experiments running. A few ideas from that and looking at the code so far. I still need to read through the tests of forms & views.

weblab/fitting/views.py
helenst: I'm not sure why I added that check - if it gets to this call then `self.runnable` will have been assigned in `form_valid`. I'll remove it.

weblab/fitting/views.py
helenst: You're right - it's only when using `FormMixin` that this logic isn't provided. `FormView` does. I think it makes sense to follow the default Django logic in this case and just implement `form_valid` handling. I'll remove the `post` implementation.

helenst:
TODO: add tests for fitting experiment / dataset support in `check_access_token`
- [ ] add `check_access_token` test cases

jonc125:
I've just noticed a slight oddity with the drop-downs: once you have select one model (say) you then need to select the '----' entry before you can click again to see the full list of models!

weblab/fitting/views.py
helenst: No reason, just an oversight! :)

helenst:
> I've just noticed a slight oddity with the drop-downs: once you have select one model (say) you then need to select the '----' entry before you can click again to see the full list of models!

Yeah, this arises because of the restrictions that various other fields place on one another (mostly protocols, fitting specs and datasets) - I may need to revisit the way this logic works as it's not very user friendly.

weblab/fitting/views.py
```diff
@@ -218,24 +218,51 @@ class FittingResultCreateView(LoginRequiredMixin, PermissionRequiredMixin, UserF
...
+            initial['model_version'] = get_object_or_404(
+                CachedModelVersion.objects.visible_to_user(self.request.user),
+                pk=model_version_id)
+            initial['model'] = initial['model_version'].model
+        elif model_id:
+            initial['model'] = get_object_or_404(
+                ModelEntity.objects.visible_to_user(self.request.user),
+                pk=model_id)
+
+        elif protocol_version_id:
```
jonc125, line 239: I'd be inclined to make this an `if` not an `elif`. We might not supply both model & protocol in normal usage, but why prevent future use when we don't need to? Similarly for `fittingspec_version_id` and `dataset_id`.

weblab/templates/fitting/fittingresultversion_detail.html
```diff
@@ -48,6 +48,8 @@ <h2 id="entityversionname">
         <a href="{% url 'entities:version' 'model' fittingresult.model.pk fittingresult.model_version.sha %}">{{ fittingresult.model.name }} @ {{ fittingresult.nice_model_version }}</a>
           &amp; protocol:
           <a href="{% url 'entities:version' 'protocol' fittingresult.protocol.pk fittingresult.protocol_version.sha %}">{{ fittingresult.protocol.name }} @ {{ fittingresult.nice_protocol_version }}</a>
+          &amp; fitting spec:
+          <a href="{% url 'fitting:version' 'spec' fittingresult.fittingspec.pk fittingresult.protocol_version.sha %}">{{ fittingresult.fittingspec.name }} @ {{ fittingresult.nice_fittingspec_version }}</a>
```
jonc125, line 52:
```suggestion
          <a href="{% url 'fitting:version' 'spec' fittingresult.fittingspec.pk fittingresult.fittingspec_version.sha %}">{{ fittingresult.fittingspec.name }} @ {{ fittingresult.nice_fittingspec_version }}</a>
```

helenst:
OK, thinking some more about how the protocols, datasets and fitting specs restrict one another:

Say we select the protocol first, then that will restrict datasets and fitting specs to those connected to that protocol.
Currently it also restricts the protocol choice to just that protocol, and you have to deselect the protocol to get the full list back. We could have it keep all protocols available, but...

If you select the protocol as above, and then select the dataset, selecting the dataset will still restrict the protocol to just the selected protocol, because the two fields mutually control one another - both get locked in to one another until you fully deselect one of them. I guess that's a side effect of having the restrictions working in both directions. It's similar for any pairing of protocol, dataset, fitting spec being 

I guess one answer might be to keep some sense of the order in which things should be selected - e.g. protocol -> datasets -> fitting specs  - but that might be a bit restrictive for the user. Or maybe have a "clear selections" button to reset all the dropdowns (except those preselected on page load).

Or perhaps the dropdowns should keep the "invalid" options for those fields, indicating that they are so, but also allowing those options to be selected, which would then trigger a reset of values in the other dropdowns.

jonc125:
I think possibly a combination of measures? I like the idea of a 'clear selections' button that resets everything as you suggest.

As for the restrictions, I think making a selection *should* restrict all the *other* drop-downs, but not the one you just selected.

So e.g. selecting a protocol would leave the other options in the menu, *until* a dataset was then selected, since that would restrict the protocol field. Similarly if they started with choosing a dataset, the protocol menu would be restricted but not the dataset menu (until a protocol was chosen).

Does that make sense?

In normal use you probably wouldn't see it happening. It only gets odd really in the case where you accidentally click on the wrong entry, so you click the menu to change it and discover your choices have disappeared!

helenst:
@jonc125 I'm done adding to this for now so feel free to re-review :)

weblab/fitting/tests/test_views.py
```diff
@@ -14,6 +17,17 @@
 from repocache.populate import populate_entity_cache
 
 
+@pytest.fixture
+def fits_user(logged_in_user):
```
jonc125, line 21: I think this is in `conftest.py` now?

jonc125 (approved):
Looks good!

PR #294, opened 2020-09-30
225 transfer ownership

sroderick-g5sro:
Fixes #294


weblab/entities/views.py
```diff
@@ -778,6 +781,54 @@ def get(self, request, *args, **kwargs):
...
+        if form.is_valid():
+            email = form.cleaned_data['email']
+            entity = self.get_object()
+            user = User.objects.filter(email=email).first()
+
+            if user is None:
+                form.add_error(None, "Unknown user")
+                return self.form_invalid(form)
+
+            if self.model.objects.filter(name=entity.name, author=user).exists():
```
sroderick-g5sro, line 814: @jonc125 I wonder if you could help me with this line of code please. It seems to return true even when the other user doesn't have an entity of the same name.

weblab/entities/views.py
```diff
@@ -15,6 +15,8 @@
     PermissionRequiredMixin,
     UserPassesTestMixin,
 )
+
+
```
jonc125, line 19: I think these were added in error?

weblab/entities/views.py
```diff
@@ -51,9 +53,10 @@
     FileUploadForm,
     ModelEntityForm,
     ProtocolEntityForm,
-)
+    EntityTransferForm)
```
jonc125, line 56: As a code style thing, it's best practice to keep the closing bracket on a line by itself, and add the extra import on its own line with a trailing comma, in alphabetical order.

weblab/entities/tests/test_views.py
```diff
@@ -842,6 +842,42 @@ def test_lists_my_protocols(self, client, logged_in_user):
...
+
+        assert model.author.email == 'test@example.com'
+        response = client.post(
+            '/entities/models/%d/transfer' % model.pk,
+            data={
+                'email': other_user.email,
+            },
+            )
+        assert response.status_code == 302
+        model2 = ModelEntity.objects.first()
```
jonc125, line 860: You should be able to do `model.refresh_from_db()` and test fields on `model` rather than creating a `model2` variable.

weblab/entities/tests/test_views.py
```diff
@@ -842,6 +842,42 @@ def test_lists_my_protocols(self, client, logged_in_user):
...
+        assert model2.author.email == 'other@example.com'
```
jonc125, line 861: I'd check `assert model.author == other_user` to make the connection more explicit.

weblab/entities/tests/test_views.py
```diff
@@ -842,6 +842,42 @@ def test_lists_my_protocols(self, client, logged_in_user):
...
+        assert model2.repocache.latest_version.sha == commit.sha
```
jonc125, line 862: As mentioned, we should also check the repo_abs_path etc., but only once that part is implemented.

weblab/entities/views.py
jonc125: I can't see any errors here, and indeed the `test_transfer` is passing on Travis, which suggests it's working as intended? The build failure is just due to flake8 & isort errors - you might want to look at installing and running those locally before commit/push?

weblab/templates/entities/entity_transfer_ownership.html
```diff
@@ -0,0 +1,17 @@
+{% extends "base.html" %}
+{% load entities %}
+
+{% block title %}{{ entity.display_type|capfirst }} rename {% endblock title %}
```
jonc125, line 4:
```suggestion
{% block title %}Change owner for {{ entity.display_type }}{% endblock title %}
```

weblab/entities/views.py
```diff
@@ -778,6 +780,54 @@ def get(self, request, *args, **kwargs):
...
+        form = self.get_form()
+
+        if form.is_valid():
+            email = form.cleaned_data['email']
+            entity = self.get_object()
+            user = User.objects.filter(email=email).first()
+
+            if user is None:
+                form.add_error(None, "Unknown user")
+                return self.form_invalid(form)
```
jonc125, line 811: It might be nice to do this in the same was as for `EntityCollaboratorForm`, i.e. add the same method `clean_email` to the form. It sets `self.cleaned_data['user']` if the user exists, or marks the form invalid if not. Then here we can just say `user = form.cleaned_data['user']`.

sroderick-g5sro: thanks, I've put this in, it works for valid email addresses but for invalid ones I get an error 

_The view entities.views.TransferView didn't return an HttpResponse object. It returned None instead._

When you say the clean_email method "marks the form invalid if not" do I need to do this explicitly or is  `raise ValidationError('User not found')` sufficient?

jonc125: The `raise` does that. I think the error is because you don't have an else branch in your view on the `if form.is_valid()`?

sroderick-g5sro: yes that did the trick thanks, I put `return self.form_invalid(form)` in the else branch and the page now works as expected

sroderick-g5sro:
@jonc125 please could you review what I have done so far, there is a failing test for datasets. It is complaining that it can't find the archive file.

weblab/datasets/views.py
```diff
@@ -230,3 +231,48 @@ def test_func(self):
...
+            if self.model.objects.filter(name=dataset.name, author=user).exists():
+                form.add_error(None, "User already has a dataset called %s" % dataset.name)
+                return self.form_invalid(form)
+
+            old_path = dataset.archive_path
+            dataset.author = user
+            dataset.save()
+            new_path = dataset.archive_path
+            if old_path.exists():
+                shutil.move(str(old_path), str(new_path))
```
jonc125, line 270: I've figured it out! The error is actually complaining that the *destination folder* doesn't exist, not the source file. So we need to create that folder first (and indeed delete the source folder after doing the move) using the same approach as in `datasets/signals.py` for when a dataset is created or deleted:
```suggestion
            dataset.abs_path.mkdir(exist_ok=False, parents=True)
            if old_path.exists():
                shutil.move(str(old_path), str(new_path))
            if old_path.parent.is_dir():
                shutil.rmtree(str(old_path.parent))
```

weblab/datasets/tests/test_views.py
```diff
@@ -194,6 +194,27 @@ def test_admin_can_delete_dataset(
...
+        response = client.post(
+            '/datasets/%d/transfer' % dataset.pk,
+            data={
+                'email': other_user.email,
+            },
+        )
+        assert response.status_code == 302
+        dataset.refresh_from_db()
+        assert dataset.author == other_user
+        assert not oldpath.exists()
```
jonc125, line 214: Based on the bug fix, I suggest we also test `assert not oldpath.parent.exists()`.

weblab/datasets/tests/test_views.py
```diff
@@ -194,6 +194,27 @@ def test_admin_can_delete_dataset(
         assert not dataset.archive_path.exists()
 
 
+@pytest.mark.django_db
+class TestTransferDatasetView:
+
+    def test_datasets_transfer_success(self, client, logged_in_user, other_user, my_dataset_with_file, helpers):
```
jonc125, line 200: As well as this success case, we should test that transfer works for a dataset without files (`my_dataset` fixture), and for error if a non-existent user, or user with a same named dataset, are given as the target.

jonc125 (changes requested):
I managed to find the bug! A few other thoughts while reading through the latest changes too.

weblab/datasets/forms.py
```diff
@@ -50,3 +51,26 @@ class DatasetFileUploadForm(forms.ModelForm):
     class Meta:
         model = DatasetFile
         fields = ['upload']
+
+
+class DatasetTransferForm(UserKwargModelFormMixin, forms.Form):
```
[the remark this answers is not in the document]
sroderick-g5sro: Yes I believe it can but I'm not sure of the correct syntax. Should I be referencing the `EntityTransferForm` in` dataset\views.py ` or do I need a sub form in `dataset\forms.py`. I can only find examples of sub forms for model forms.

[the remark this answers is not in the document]
jonc125: I think in `datasets/views.py` you can use `from entities.forms import EntityTransferForm` and then use that where currently you have `DatasetTransferForm`.

weblab/datasets/urls.py
```diff
@@ -5,7 +5,6 @@
 
 _FILENAME = r'(?P<filename>[\w\-. \%:]+)'
 _FILEVIEW = r'%s/(?P<viz>\w+)' % _FILENAME
-
```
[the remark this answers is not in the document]
sroderick-g5sro: I was a bit confused by the - is that a git thing?

weblab/datasets/forms.py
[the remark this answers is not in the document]
jonc125: Though given that a dataset is not an entity, perhaps it makes sense to move that class into `accounts/forms.py` and name it `OwnershipTransferForm`?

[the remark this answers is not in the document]
jonc125: Oh, and it doesn't need to inherit from `UserKwargModelFormMixin` I think, as you don't make use of the logged in user info, only the manually entered email address.

weblab/datasets/urls.py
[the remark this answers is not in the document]
jonc125: That's the diff marker for a deleted line. I'm suggesting that we *do* want a blank line after the assignment to `_FILEVIEW`.

weblab/datasets/forms.py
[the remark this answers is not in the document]
sroderick-g5sro: interestingly when I remove `UserKwargModelFormMixin` I get the following error `__init__() got an unexpected keyword argument 'user'`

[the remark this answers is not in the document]
jonc125: Ah, you also need to remove the corresponding base class from the view(s) which provides that argument. See https://django-braces.readthedocs.io/en/latest/form.html#userformkwargsmixin for how this works.

weblab/accounts/forms.py
```diff
@@ -1,5 +1,7 @@
+from braces.forms import UserKwargModelFormMixin
```
jonc125, line 1: This isn't needed.

weblab/datasets/forms.py
```diff
@@ -52,6 +53,29 @@ class Meta:
         fields = ['upload']
 
 
+class DatasetTransferForm(UserKwargModelFormMixin, forms.Form):
```
jonc125, line 56: You can delete this class here (and the import of `User` above).

weblab/datasets/views.py
```diff
@@ -24,12 +25,16 @@
 from core.combine import ManifestWriter
 from core.visibility import VisibilityMixin
 
+
+from .forms import DatasetAddFilesForm, DatasetFileUploadForm, DatasetForm, 
 from .forms import (
     DatasetAddFilesForm,
     DatasetFileUploadForm,
     DatasetForm,
+    DatasetTransferForm,
```
jonc125, line 34: Remove this line, and add above line 25
```python
from accounts.forms import OwnershipTransferForm
```

weblab/datasets/views.py
```diff
@@ -24,12 +25,16 @@
...
     DatasetRenameForm,
 )
+
```
jonc125, line 37: Remove blank line here.

weblab/entities/views.py
```diff
@@ -42,6 +42,7 @@
 from fitting.models import FittingSpec
 from repocache.exceptions import RepoCacheMiss
 from repocache.models import CachedProtocolVersion
+from accounts.forms import OwnershipTransferForm
```
jonc125, line 45: This line will need to move up to be alphabetical.

jonc125 (commented):
I've had a quick look through...

weblab/entities/views.py
```diff
@@ -794,6 +795,49 @@ def get(self, request, *args, **kwargs):
...
+            entity = self.get_object()
+            if self.model.objects.filter(name=entity.name, author=user).exists():
+                form.add_error(None, "User already has a %s called %s" % (self.model.display_type, entity.name))
+                return self.form_invalid(form)
+
+            old_path = entity.repo_abs_path
+            entity.author = user
+            entity.save()
+            user.get_storage_dir('repo').mkdir(exist_ok=True, parents=True)
+            new_path = entity.repo_abs_path
```
jonc125, line 830: Looking at the latest failure, this might need to change: I think when a new spec is created it automatically creates all parent folders, whereas this misses the last 'fittingspec' segment. Instead we could do:
```suggestion
            new_path = entity.repo_abs_path
            new_path.parent.mkdir(exist_ok=True, parents=True)
```
which makes sure all levels above where we're moving to are created.

weblab/entities/views.py
```diff
@@ -794,6 +795,50 @@ def get(self, request, *args, **kwargs):
...
+            if self.model.objects.filter(name=entity.name, author=user).exists():
+                form.add_error(None, "User already has a %s called %s" % (self.model.display_type, entity.name))
+                return self.form_invalid(form)
+
+            old_path = entity.repo_abs_path
+            entity.author = user
+            entity.save()
+            new_path = entity.repo_abs_path
+            new_path.parent.mkdir(exist_ok=True, parents=True)
+
```
jonc125, line 831:
```suggestion
```

jonc125 (approved):
Looks good!

PR #298, opened 2020-10-08
Add cancel button to entity delete template

### LINKED ISSUE #239: User experience enhancements in opening and manipulating entities
During user testing a number of minor enhancements were identified that might help to improve a user's experience. These are listed here:

1) When selecting delete on any entity, dataset or fitting spec - the user is asked to confirm deletion. But there is no cancel button. The cancel button should return the user to their former location

2) When selecting a given entity there is a flash of "unformatted content". This is due to delays in loading dynamic data on the js pages. If some means of delaying loading until the data was complete can be found this may make the page loading a smoother experience.

3) Currently compare actions are selected using the compare button. The button increases in size and presents the user with mutiple links. Use of a drop down and compare button may may navigation slightly more intuitive. 

4) Selecting "compare selected experiments" when there are no checkboxes selected results in "page not found". This might be enhanced by not allowing selection of the button unless checkboxes are selected or alternatively display of a custom error message

sroderick-g5sro:
Fixes the first item in #239

### DIFF
--- a/weblab/templates/entities/entity_confirm_delete.html
+++ b/weblab/templates/entities/entity_confirm_delete.html
@@ -8,6 +8,7 @@
     <p>Are you sure you want to delete all versions of {{ object.display_type }} "{{ object }}"?</p>
     <p><strong>This operation cannot be undone.</strong></p>
     <button>Confirm</button>
+    <a class="button" href="{% entity_url 'detail' object %}" title="Do not delete">Cancel</a>
 </form>
 
 {% endblock %}
--- a/weblab/templates/entities/entity_confirm_delete.html
+++ b/weblab/templates/entities/entity_confirm_delete.html
@@ -1,11 +1,11 @@
 {% extends "base.html" %}
 
-{% block title %}Delete {{ object.display_type }} - {% endblock title %}
+{% block title %}Delete entity{% endblock title %}
 
 {% block content %}
 
 <form action="" method="post">{% csrf_token %}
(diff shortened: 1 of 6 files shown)


weblab/templates/entities/entity_confirm_delete.html
```diff
@@ -8,6 +8,7 @@
     <p>Are you sure you want to delete all versions of {{ object.display_type }} "{{ object }}"?</p>
     <p><strong>This operation cannot be undone.</strong></p>
     <button>Confirm</button>
+    <a class="button" href="{% entity_url 'detail' object %}" title="Do not delete">Cancel</a>
```
sroderick-g5sro, line 11: @jonc125 I tried this as per the note on the issue but it errors with 'Entity' object has no attribute 'url_type'

jonc125: Hah, good point! I hadn't anticipated that because we use `display_type` above, but actually that gives an empty string because it's not set for the generic `Entity` base class either. So maybe `href="javascript:history.back()"` would be better? Even if it does cheat a little!

jonc125: And we should probably get rid of the `display_type` reference too.

jonc125 (changes requested):
These look good. You should also do experiment & experimentversion.

jonc125 (approved):

PR #300, opened 2020-10-12
Lists of fitting results

### LINKED ISSUE #276: Simple list views of fitting results
The simplest way to navigate to and view results of individual fitting experiments is just to list everything linked to a given model/protocol/dataset/fitting spec version.

There's commented out code in `entity_version.html` for a button that could link to such a view. Though the button text should probably read 'Fitting results'.

The view itself will be very similar to `entities/compare_experiments.html`, except that initially it won't have the comparison checkboxes, just the individual links. (Until the comparison views from #134 are done.)

- [x] List all fitting results for a given model version
    - Group the list by dataset
- [x] List all fitting results for a given dataset
    - Group the list by model
    - Have a button to show all versions or only the latest
- [x] List all fitting results for a given protocol
    - Group by dataset, subgroup by model
    - Have a button to show all model versions or only the latest
- [x] List all fitting results for a given fitting spec
    - Group by dataset, subgroup by model
    - Have a button to show all model versions or only the latest

Child of #135.

helenst:
Fixes #276

### DIFF
--- a/weblab/datasets/tests/test_views.py
+++ b/weblab/datasets/tests/test_views.py
@@ -637,3 +637,96 @@ class TestDatasetVisibility:
     def test_nonexistent_dataset_generates_404_for_user(self, client, logged_in_user, helpers, recipe, url):
         response = client.get(url % 10000)
         assert response.status_code == 404
+
+
+@pytest.mark.django_db
+class TestDatasetCompareFittingResultsView:
+    def test_shows_fittings_related_to_dataset(self, client, fittingresult_version):
+        fit = fittingresult_version.fittingresult
+
+        # should not be included, as it uses a different dataset
+        recipes.fittingresult_version.make()
+
+        response = client.get('/datasets/%d/fittings' % fit.dataset.pk)
+
+        assert response.status_code == 200
+        assert response.context['comparisons'] == [(fit.model, [fit])]
+
+    def test_groups_by_model(self, client, helpers, public_dataset):
+        m1, m2 = recipes.model.make(_quantity=2)
+        m1v = helpers.add_cached_version(m1, visibility='public')
+        m2v = helpers.add_cached_version(m2, visibility='public')
+
+        # Create publicly visible fitting result versions
+        fit1_m1 = recipes.fittingresult_version.make(
+            fittingresult__dataset=public_dataset,
+            fittingresult__model=m1,
+            fittingresult__model_version=m1v,
+            fittingresult__fittingspec_version__visibility='public',
+            fittingresult__protocol_version__visibility='public',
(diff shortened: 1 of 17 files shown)


weblab/templates/datasets/compare_fittings.html
```diff
@@ -0,0 +1,78 @@
+{% extends "base.html" %}
+{% load entities %}
+{% load fittings %}
+{% load staticfiles %}
+
+{% block title %}Dataset : compare fitting experiments - {% endblock title %}
+
+{% block content %}
```
jonc125, line 8: There's a lot of overlap here with the dataset_detail template. Could we make a new dataset_base or similar that both inherit from maybe?

weblab/entities/tests/test_views.py
```diff
@@ -652,6 +652,290 @@ def test_returns_404_if_commit_not_found(self, client, logged_in_user):
...
+            '/entities/models/%d/versions/%s/fittings' % (fit.model.pk, fit.model_version.sha)
+        )
+
+        assert response.status_code == 200
+        assert response.context['comparisons'] == [(fit.dataset, [fit])]
+
+    def test_groups_by_dataset_for_model(self, client, public_model):
+        model_version = public_model.repocache.latest_version
+        dataset1, dataset2 = recipes.dataset.make(_quantity=2, visibility='public')
+
```
jonc125, line 676: Similarly to above, can we add some private versions and check the corresponding results aren't included?

jonc125 (changes requested):
Looking good. I tweaked some aspects of the display; figured it was easier just to make the changes than put them as review comments. There are just a few minor things to finish off I think.

jonc125 (approved):
Just a very minor suggestion!

I was confused at first that the visibility form still wasn't showing for me (hence the extra test added) but killing & restarting my dev server resolved it 🤷‍♂️

weblab/datasets/tests/test_views.py
```diff
@@ -730,3 +730,41 @@ def test_multiple_model_versions_for_dataset(self, client, helpers, public_datas
...
             (m2, [fit3_m2v]),
         ]
+
+    def test_ensure_private_results_are_not_shown(self, client, public_dataset):
+        recipes.fittingresult_version.make(
+            fittingresult__dataset=public_dataset,
+            fittingresult__model_version__visibility='private',
+            fittingresult__protocol_version__visibility='public',
+            fittingresult__fittingspec_version__visibility='public',
+        ).fittingresult
```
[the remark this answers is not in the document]
helenst: Indeed - have fixed this, and similarly in entity view tests.

PR #302, opened 2020-10-14
Rerun experiments enhancement

### LINKED ISSUE #230: Rerun experiments re-runs on all previous versions of model/protocol instead of just latest
When uploading a new version of a model/protocol the 're-run experiments using this version' is really nice, but it is re-running all the old protocols/models and not just the latest one, which is probably what we want the default behaviour to be.

sroderick-g5sro:
Fixes #230

### DIFF
--- a/weblab/entities/processing.py
+++ b/weblab/entities/processing.py
@@ -160,8 +160,10 @@ def process_check_protocol_callback(data):
 def record_experiments_to_run(user, entity, commit):
     """Record what experiments to run automatically on a new entity version.
 
-    The latest experiments that were run with the parent version(s) (that are visible to the user)
-    should be repeated with the new version.
+    Find all experiments run with the entity for which we're adding a new version (consider case of a model)
+    Get the list of corresponding protocol IDs. We now know both our model ID, and all protocols that have
+    previously had any version run on any version of this model.
+    Run our new model version under the latest (visible) version of all those protocols.
 
     :param user: the user that created the new version
     :param entity: the entity that has had a new version created
--- a/weblab/entities/processing.py
+++ b/weblab/entities/processing.py
@@ -9,8 +9,7 @@ from django.db import IntegrityError
 from experiments.models import Experiment, PlannedExperiment
 from repocache.models import ProtocolInterface, ProtocolIoputs
 
-from .models import AnalysisTask
-
+from .models import AnalysisTask, ModelEntity, ProtocolEntity
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


weblab/entities/processing.py
```diff
@@ -171,21 +170,24 @@ def record_experiments_to_run(user, entity, commit):
...
+            if other_entity.is_version_visible_to_user(other_version.sha, user):
                 # Record the new experiment to run
                 kwargs = {
                     'submitter': user,
-                    'model_id': expt.model_id,
-                    'model_version': expt.model_version.sha,
-                    'protocol_id': expt.protocol_id,
-                    'protocol_version': expt.protocol_version.sha,
+                    entity.other_type: other_entity,
+                    other_entity + '_version': other_version.sha,
```
sroderick-g5sro, line 189: This line errors with `TypeError: unsupported operand type(s) for +: 'ProtocolEntity' and 'str'`. I wonder if the fix is as simple as wrapping other_entity in str()?

jonc125: Ah, no, I think it should be `entity.other_type + '_version'` sorry!

weblab/entities/processing.py
```diff
@@ -160,7 +160,7 @@ def process_check_protocol_callback(data):
 def record_experiments_to_run(user, entity, commit):
     """Record what experiments to run automatically on a new entity version.
 
-    Any experiments that were run with the parent version(s) (that are visible to the user)
+    The latest experiments that were run with the parent version(s) (that are visible to the user)
```
jonc125, line 163: I think it's worth expanding this description slightly to be more detailed, based on what's in https://github.com/ModellingWebLab/WebLab/issues/230#issuecomment-705484207. Would you agree?

weblab/entities/tests/test_views.py
```diff
@@ -1418,22 +1419,24 @@ def _add_experiment(proto_author, proto_vis, shared=False,
             }), content_type='application/json')
             assert response.status_code == 200
             detail = json.loads(response.content.decode())
+            print(detail)
```
jonc125, line 1422:
```suggestion
```

weblab/entities/tests/test_views.py
```diff
@@ -1418,22 +1419,24 @@ def _add_experiment(proto_author, proto_vis, shared=False,
...
             assert 'updateEntityFile' in detail
+
```
jonc125, line 1424:
```suggestion
```

jonc125 (approved):
Looks good! I've suggested a few minor style/documentation tweaks. No need to get a re-review once those are done though.

PR #305, opened 2020-10-16
More robust upgrade process

jonc125:
Extra robustness for #287 - discovered to be needed running an upgrade on a cloud staging server.


jonc125:
I've now managed to run an upgrade deployment successfully on Azure, so am happy I won't break the main server!

sroderick-g5sro (approved):

PR #306, opened 2020-10-19
Handle simultaneous updates gracefully

### LINKED ISSUE #45: Ensure simultaneous attempts at updating entities are handled gracefully
Following on from work in #21 and #44, ensure that if users A and B are attempting to modify the same entity, the two modifications do not interfere with one another.

This should probably involve presenting user B with an error message on the new version form, indicating that the entity was updated by user A since they loaded the page. Invite them to try again, applying their changes on top of user A's version.

Dealing with any kind of merge or conflict resolution could get fiddly from a UI perspective, so probably best to keep it simple for now and just prevent the repo getting into a bad state, warning the user so they can try again.

sroderick-g5sro:
Fixes #45

### DIFF
--- a/weblab/entities/tests/test_views.py
+++ b/weblab/entities/tests/test_views.py
@@ -1417,7 +1417,7 @@ class TestVersionCreation:
             response = client.post(
                 '/entities/models/%d/versions/new' % model.pk,
                 data={
-                    'parent_hexsha': m1.repo.latest_commit.sha,
+                    'parent_hexsha': model.repo.latest_commit.sha,
                     'filename[]': ['uploads/file2.txt'],
                     'mainEntry': ['file2.txt'],
                     'commit_message': 'new',
--- a/weblab/entities/tests/test_views.py
+++ b/weblab/entities/tests/test_views.py
@@ -96,6 +96,26 @@ class TestEntityCreation:
         assert response.status_code == 302
         assert '/login/' in response.url
 
+    def test_error_if_later_version(self, logged_in_user, client, helpers):
+        helpers.add_permission(logged_in_user, 'create_model')
+        model = recipes.model.make(author=logged_in_user)
+        helpers.add_version(model, tag_name='v1')
+
+        response = client.post(
+            '/entities/models/%d/versions/new' % model.pk,
+            data={
+                'parent_hexsha': '',
+                'filename[]': ['file1.txt'],
+                'delete_filename[]': [],
+                'mainEntry': ['file1.txt'],
+                'commit_message': 'files',
+                'tag': 'v1',
+                'visibility': 'public',
+            },
+        )
+
+        assert response.status_code == 200
+
 
 @pytest.mark.django_db
 class TestEntityRenaming:
--- a/weblab/entities/tests/test_views.py
+++ b/weblab/entities/tests/test_views.py
@@ -117,6 +117,7 @@ class TestEntityCreation:
         assert response.status_code == 200
         assert "Someone has saved a newer version since you started " in response.rendered_content
 
+
 @pytest.mark.django_db
 class TestEntityRenaming:
     def test_model_renaming_success(self, client, logged_in_user, helpers):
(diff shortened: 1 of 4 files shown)


weblab/entities/tests/test_views.py
```diff
@@ -96,6 +96,26 @@ def test_create_protocol_requires_permissions(self, logged_in_user, client):
...
+                'filename[]': ['file1.txt'],
+                'delete_filename[]': [],
+                'mainEntry': ['file1.txt'],
+                'commit_message': 'files',
+                'tag': 'v1',
+                'visibility': 'public',
+            },
+        )
+
+        assert response.status_code == 200
```
jonc125, line 117: Looks good. It would be nice to check that we have an expected error message too with this kind of thing.

sroderick-g5sro: I have done this using `assert "Someone has saved a newer version since you started " in response.rendered_content`. Not sure if there is a better way? Also do I need to check anything else after the post please?

sroderick-g5sro:
I'm confused by the travis failure. Its runnning different code for the branch for `test_rerun_experiments`

jonc125:
> I'm confused by the travis failure. Its runnning different code for the branch for `test_rerun_experiments`

It's the PR build rather than the branch build that's failing. While the branch build tests the code as it exists on this branch, the PR build tests what the result would be of merging this branch into master. So the failing test looks to exist more on master rather than this branch. Which is odd because the latest commit on master claims to have passed all tests...

jonc125:
Ah, I see what the issue is. You've added a line in this PR setting the parent_hexsha, but the variable you set it from has been renamed on master so the auto-merge results in a failing test. You'll need to manually merge master into this branch and then use the new variable name.

sroderick-g5sro:
thanks @jonc125, how do I go about merging master into this branch?

weblab/entities/tests/test_views.py
jonc125: That does work. I thought it might be the best option, then I realised that `FormMixin` provides `form` as a variable in the template context for us, so we can use that to test errors more programmatically. Something like
```
assert None in response.context['form'].errors
assert 'Someone has saved a newer version' in response.context['form'].errors[None]
```
I think? I don't think you need to test anything beyond that the error exists, although if you wanted to be paranoid I guess you could check that the latest commit in the repo hasn't changed?

jonc125:
> thanks @jonc125, how do I go about merging master into this branch?

I think it has to be done from your local git client, rather than via GitHub. So (if using the command line):
```sh
git checkout master
git pull  # Ensure local master is up-to-date with GitHub
git checkout 45-handle-simultaneous-updates
git merge master
git push
```

weblab/entities/tests/test_views.py
sroderick-g5sro: Thanks Jonathan, I couldn't get your two lines of code to run. Do I need to specify FormMixin on the test method?

sroderick-g5sro:
> > thanks @jonc125, how do I go about merging master into this branch?
> 
> I think it has to be done from your local git client, rather than via GitHub. So (if using the command line):
> 
> ```shell
> git checkout master
> git pull  # Ensure local master is up-to-date with GitHub
> git checkout 45-handle-simultaneous-updates
> git merge master
> git push
> ```

Works perfectly. Thank you :)

weblab/entities/tests/test_views.py
jonc125: You shouldn't do, because it's included in `EntityNewVersionView` which this is testing. Maybe print `response.context` to see what is there?

sroderick-g5sro: Thanks. I have done a search on response.context and can't see anywhere where the error messages appear.

jonc125: Does the form appear at all?

sroderick-g5sro: I don't believe so

jonc125: Oh well! Let's stick with your existing approach then :)

jonc125 (approved):
Looks good!

PR #308, opened 2020-10-23
Dont allow adding self as a collaborator

### LINKED ISSUE #127: Don't allow adding yourself as a collaborator!
It doesn't have any real effect, but certainly appears odd!

sroderick-g5sro:
Fixes #127

### DIFF
--- a/weblab/entities/forms.py
+++ b/weblab/entities/forms.py
@@ -151,7 +151,8 @@ class EntityCollaboratorForm(forms.Form):
         user = self._get_user(email)
         if not user:
             raise ValidationError('User not found')
-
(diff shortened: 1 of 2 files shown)


jonc125 (changes requested):
This looks fine, but we should add a test that would fail without this change, in entities test_forms.py I think, within `TestEntityCollaboratorFormSet`. It would be like `test_add-collaborator` but rather than using the `public_model` fixture you can create a model owned by `model_creator`:
```python
model = recipes.model.make(author=model_creator)
```
And `assert not form.is_valid()` of course :)

jonc125 (approved):
Great!

PR #310, opened 2020-10-27
Changed tag validation

### LINKED ISSUE #309: Tag entry still allowing invalid characters
Tag validation errors with Git command error: stderr: 'fatal: . Regex expression incorrect need to use r'^[-_A-Za-z0-9]+$' instead

sroderick-g5sro:
Fixes #309

### DIFF
--- a/weblab/entities/views.py
+++ b/weblab/entities/views.py
@@ -666,7 +666,7 @@ class EntityNewVersionView(
                     entity.add_tag(tag, commit.sha)
                 except GitCommandError as e:
                     entity.repo.rollback()
-                    return self.fail_with_git_errors([e.stderr])
+                    return self.fail_with_git_errors([e.stderr], form)
 
             # Temporary upload files have been safely committed, so can be deleted
             for filepath in files_to_delete:
@@ -685,7 +685,7 @@ class EntityNewVersionView(
                 reverse(ns + ':version', args=[entity.url_type, entity.id, commit.sha]))
         else:
             # Nothing changed, so inform the user and do nothing else.
-            form = self.get_form()
+
             form.add_error(None, 'No changes were made for this version')
             return self.form_invalid(form)
 
--- a/weblab/entities/views.py
+++ b/weblab/entities/views.py
@@ -596,11 +596,11 @@ class EntityNewVersionView(
         return super().get_context_data(**kwargs)
 
     def post(self, request, *args, **kwargs):
+
+        entity = self.object = self.get_object()
         form = self.get_form()
         if not form.is_valid():
             return self.form_invalid(form)
-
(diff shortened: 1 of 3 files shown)


jonc125 (changes requested):
This is the correct fix to the regex. However it also needs changing in `EntityTagVersionForm` below, and we should probably add a couple of tests that fail with the old regex to prove that it's working! They should both live in entities test_views.py. The one for this case would need to be a new method in the `TestVersionCreation` class, and for the other form in `TestTagging`. Basically you want to test that if you post with an invalid tag in each case, you get a 200 response with the error message showing.

jonc125:
Ah, good point - you've reminded me that the `post` method in `EntityNewVersionView` doesn't ever call `form.is_valid` to check for form validation errors! (Because until we added this, there weren't any, so all the validation was done within `post` itself.) So the first thing `post` should now do is probably something like this:
```python
form = self.get_form()
if not form.is_valid():
    return self.form_invalid(form)
```

Given that we'll then have a reference to the form, we probably shouldn't call `get_form` again for the same request, so should remove later calls, and also pass `form` to the `fail_with_git_errors` method.

weblab/entities/views.py
```diff
@@ -685,8 +689,8 @@ def post(self, request, *args, **kwargs):
             form.add_error(None, 'No changes were made for this version')
             return self.form_invalid(form)
 
-    def fail_with_git_errors(self, git_errors):
-        form = self.get_form()
+    def fail_with_git_errors(self, git_errors, form):
```
jonc125, line 692: This method is called in another place too which doesn't pass the `form` argument - interestingly because we don't see the invalid tag error any more! We might be able to trigger it if we add a test with a duplicate (but otherwise valid) tag added to a new version.

Also 4 lines up we have another `form = self.get_form()` line that should be removed.

jonc125 (changes requested):
This is looking good; just a couple of bits of nit-picking I noticed!

weblab/entities/views.py
sroderick-g5sro: Thanks Jonathan, sorry I'm not quite clear about the first part of this, where you say we dont see the invalid tag error any more. I thought I was checking for this in the rendered_content?

jonc125: It's on [line 665](https://github.com/ModellingWebLab/WebLab/pull/310/files#diff-4bef1f458b6904ff9e9b2d64655045de1bc3c755be9d9f7968a972239941b85aL665) in that file, where a git command line error adding the tag is caught. Since we check for invalid tags in the form, we're no longer getting errors when calling git!

jonc125 (approved):
Great!

PR #312, opened 2020-10-28
Upgrade to django 2.2

### LINKED ISSUE #192: Add option to delete account
https://scrambler.cs.ox.ac.uk/privacy

> We will be adding the facility to delete your account shortly.

### LINKED ISSUE #195: Upgrade to Django-2
I accidentally had django-2 installed and found that code is not compatible. We might want to look into this at some point.

Edit: [2.2 is now the earliest supported version](https://www.djangoproject.com/download/#supported-versions) so we'd better upgrade soon! That's also the only current LTS version.

Using `PYTHONWARNINGS=always pytest --capture=no` gives an indication of things to change in advance. We have 2 areas we currently use that are unsupported in 2.0:
```
/Users/jonc/projects/WebLab/WebLab/weblab/core/models.py:4: RemovedInDjango20Warning: Importing from django.core.urlresolvers is deprecated in favor of django.urls.
/Users/jonc/projects/WebLab/WebLab/weblab/core/models.py:31: RemovedInDjango20Warning: on_delete will be a required arg for ForeignKey in Django 2.0. Set it to models.CASCADE on models and in existing migrations if you want to maintain the current default behavior. See https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey.on_delete
```
The latter comes up in many places.

The [v2.2 changes](https://docs.djangoproject.com/en/3.1/releases/2.2/#backwards-incompatible-changes-in-2-2) don't look that scary. Likewise [v2.1](https://docs.djangoproject.com/en/3.1/releases/2.1/#backwards-incompatible-changes-in-2-1).

https://docs.djangoproject.com/en/3.1/releases/2.2/#permissions-for-proxy-models might simplify our collaborators on fitting specs issue!

To do the upgrade, we need to change the versions specified in `requirements/base.in` and run `make` within the requirements folder, then `pip install -r requirements/dev.txt`.

sroderick-g5sro:
Fixes #195


weblab/accounts/urls.py
```diff
@@ -23,3 +23,4 @@
         name='myaccount',
     ),
 ]
+app_name = 'accounts'
```
jonc125, line 26: I'm curious, as I haven't read all the docs, but what's the reason this needs to be added now?

jonc125:
> apologies there is a flake8 issue outstanding, I have tried every combination of indentation I can think of but can't resolve it

For some reason running flake8 locally wasn't picking this up (hence my messing around with upgrades!) but I think I've fixed it anyway...

weblab/accounts/urls.py
sroderick-g5sro: This is in response to the following error message 

> django.core.exceptions.ImproperlyConfigured: Specifying a namespace in include() without providing an app_name is not supported. Set the app_name attribute in the included module, or pass a 2-tuple containing the list of patterns and app_name instead

jonc125: Ah, I see. Good reason!

weblab/experiments/views.py
```diff
@@ -235,9 +235,14 @@ def get(self, request, *args, **kwargs):
             visibility_where = visibility_where | Q(entity__entity__in=visible_entities)
 
         # If specific versions have been requested, show at most those
-        q_model_versions = self.versions_query('model', model_versions, q_models, visibility_where)
-        protocol_visibility_where = visibility_where & Q(entity__entity__is_fitting_spec=False)
-        q_protocol_versions = self.versions_query('protocol', protocol_versions, q_protocols, protocol_visibility_where)
+
+        q_model_versions = self.versions_query('model', model_versions, q_models.values('pk'), visibility_where)
+        if show_fits:  # Temporary hack
```
jonc125, line 240: The merge has gone wrong here: the `show_fits` variable no longer exists in master. So you can just use the `else` clause definition for `protocol_visibility_where`.

sroderick-g5sro:
Unit tests are failing with error `psycopg2.ProgrammingError: subquery has too many columns` we previously resolved this by adding .values('pk) to q_model and q_protocol in experiments/views.py. Not sure why the error has returned?

jonc125:
I think this is coming up because you merged in Helen’s recent work on a fitting results matrix display, which added similar logic to the `fitting` app. So it you look at https://github.com/ModellingWebLab/WebLab/blob/192-upgrade-to-Django-2/weblab/fitting/views.py#L198 I reckon we need to change `q_models` to ` q_models.values('pk')` on that line. Hopefully that’s the only instance!

sroderick-g5sro:
I've made the change to add .values('pk') on line 198 but still getting the same test errors.

sroderick-g5sro:
There is one failing unit test left `test_ignores_unrelated_datasets` in `weblab/fitting/tests/test_views.py ` this only fails on python 3.5 branch build.

jonc125:
Curious, especially as the PR build passes! I've tried re-running the Travis build in case it's something intermittent. But it may be we can proceed with the merge anyway. I'll test locally tomorrow to see what happens here.

jonc125:
Well, that re-run test passed 🤷‍♂️ so no idea what went wrong the first time!

jonc125 (approved):
This all looks good, and I've done a test upgrade deployment on my local VM successfully as well. Happy to merge!

Issue #278, opened 2020-08-27
Matrix view for fitting experiments


jonc125:
Similarly to the main experiments matrix, we want a matrix view for fitting experiments.

However, for this **each fitting spec has its own matrix**, with models on the rows, and dataset+protocol combinations as the columns. (If there is only one protocol for a given dataset, you don't need to display the protocol name at all.)

@mirams @MichaelClerx how should we handle versions of models & protocols here? In the main matrix you only see the latest (visible) version of each by default (and corresponding results). For fitting would it be better to include results from earlier versions if they exist but results for the latest version(s) don't? Similarly, should we include results generated from earlier versions of the fitting spec (if there aren't corresponding results from the latest version)?

Comparison mode should be supported as for the existing matrix, so you can select multiple rows/columns/squares and compare results, and also view a sub-matrix.

Clicking on a blank square should launch the 'run fitting' view from #277, with all drop-downs pre-filled (including versions).

I'm unsure yet how much we'll be able to share JS code from the existing matrix, and how much would need to be copy & tweak.

Child of #135. Do #276 and #277 first.

MichaelClerx:
> @mirams @MichaelClerx how should we handle versions of models & protocols here? In the main matrix you only see the latest (visible) version of each by default (and corresponding results). For fitting would it be better to include results from earlier versions if they exist but results for the latest version(s) don't? Similarly, should we include results generated from earlier versions of the fitting spec (if there aren't corresponding results from the latest version)?

Not sure! I think it relates to this:

> Clicking on a blank square should launch the 'run fitting' view from #277, with all drop-downs pre-filled (including versions).

If we think seeing all the versions is important, then maybe we should let that go for fitting? Maybe clicking a square should take you to a page that shows either the most recent available results, or possibly an overview of all versions? Running could be a button on that page somewhere?

@mirams ?

jonc125:
@mirams Helen is about to start work on this, so would value your opinion on what should be shown, if you have one!

mirams:
I think you would still want a blank to show "the current version of the fitting spec hasn't been run on the current model/dataset"

jonc125:
Gary:
> Does the usual model/protocol matrix let you get to old versions of the experiment on previous versions of the model/protocol, I don't think it does? Unless I can't hit on any that have been run on old ones...
> It would certainly be nice to be able to see a list of fitting results from previous versions of the model, dataset or fitting spec...

Jonathan:
[...]

PR #314, opened 2020-10-29
Fitting results matrix

helenst:
Fitting results matrix

This is for #278 

* One per fitting spec, with models on rows and datasets on columns
* Javascript adapted to row/column rather than model/protocol so it can generically handle fitting or simulation experiments (and columns need not have versions)
* Click cell goes through to fitting results form (rather than starting an experiment directly, as with the original matrix)

I've removed the old hacky UI elements for showing fittings in a matrix, although the relevant database fields are still in place.

### DIFF
--- a/weblab/fitting/tests/test_views.py
+++ b/weblab/fitting/tests/test_views.py
@@ -469,6 +469,13 @@ class TestCreateFittingResultView:
         assert response.context['form'].initial['model'] == public_model
         assert response.context['form'].initial['model_version'] == version
 
+    def test_with_preselected_model_version_by_sha(self, client, fits_user, public_model):
+        version = public_model.repocache.latest_version
+        response = client.get('/fitting/results/new', {'model_version': version.sha})
+        assert response.status_code == 200
+        assert response.context['form'].initial['model'] == public_model
+        assert response.context['form'].initial['model_version'] == version
+
     def test_with_preselected_protocol_version(self, client, fits_user, public_protocol):
         version = public_protocol.repocache.latest_version
         response = client.get('/fitting/results/new', {'protocol_version': version.pk})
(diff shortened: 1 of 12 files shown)


jonc125 (changes requested):
Just been having a quick play and had a couple of ideas for improvement; I'll look through properly tomorrow.

1. It would be good to restrict the columns so they only show datasets linked to protocols that are linked to the fitting specification. Otherwise you get potentially many blank cells, and clicking on them leads to a fitting setup page that's "stuck" as the pre-selections aren't actually allowed!
2. The matrix page should link back to the fitting specification - perhaps in the title where its name is mentioned?

weblab/fitting/tests/test_views.py
```diff
@@ -942,3 +978,530 @@ def test_fittingspec_transfer_success(self, client, logged_in_user, other_user,
...
+        """
+        This is a dumb page that doesn't actually load any data, so we just
+        check that the URLs are working.
+        """
+        response = client.get(url % public_fittingspec.id)
+        assert response.status_code == 200
+
+    def test_matrix_page_for_fitting_spec(self, client, public_fittingspec):
+        response = client.get('/fitting/specs/%d/results' % public_fittingspec.pk)
+        assert response.status_code == 200
```
jonc125, line 1010:
```suggestion
```
Would this be a sensible reduction in number of tests?

weblab/fitting/tests/test_views.py
```diff
@@ -942,3 +978,530 @@ def test_fittingspec_transfer_success(self, client, logged_in_user, other_user,
...
+        assert response.status_code == 404
+
+
+@pytest.mark.django_db
+class TestFittingSpecResultsMatrixJsonView:
+    def test_matrix_json_for_fitting_spec(self, client, public_fittingspec):
+        response = client.get('/fitting/specs/%d/results/matrix' % public_fittingspec.pk)
+        assert response.status_code == 200
+        data = json.loads(response.content.decode())
+        assert 'getMatrix' in data
```
jonc125, line 1024: Isn't this redundant with `test_matrix_json` below?

weblab/fitting/tests/test_views.py
```diff
@@ -942,3 +978,530 @@ def test_fittingspec_transfer_success(self, client, logged_in_user, other_user,
...
+        unrelated_dataset = recipes.dataset.make(protocol=unrelated_protocol)
+
+        fittingspec = recipes.fittingspec.make(protocol=related_protocol)
+        helpers.add_version(fittingspec, visibility='public')
+
+        response = client.get('/fitting/specs/%d/results/matrix?subset=all' % fittingspec.pk)
+        data = json.loads(response.content.decode())
+
+        columns = data['getMatrix']['columns']
+        assert len(columns) == 1
```
jonc125, line 1080: Maybe make this the third assert so it's easier to see why it fails if it does?

weblab/fitting/tests/test_views.py
```diff
@@ -942,3 +978,530 @@ def test_fittingspec_transfer_success(self, client, logged_in_user, other_user,
...
+        response = client.get('/fitting/specs/%d/results/matrix?subset=all' % fit.fittingspec.pk)
+        data = json.loads(response.content.decode())
+        assert 'getMatrix' in data
+        assert len(data['getMatrix']['rows']) == 0
+        assert len(data['getMatrix']['columns']) == 1
+        assert len(data['getMatrix']['experiments']) == 0
+
+    def test_ignores_unrelated_datasets(self, client, helpers):
+        related_protocol = recipes.protocol.make()
+        helpers.add_version(related_protocol, visibility='public')
```
jonc125, line 1066: We can probably get away with `add_fake_version` here (and most/all places below?) to make the tests faster.

jonc125 (approved):
I've made a few minor suggestions, but overall I'm very happy with this - it seems to work well. So will be good to get some user feedback once deployed :)  I wonder if we should remove the 'fixes' keyword in the description before merging this so the original issue remains open to capture feedback from Michael & Gary there?

helenst:
OK - I've fixed the points you raised and removed the 'fixes' keyword @jonc125

PR #316, opened 2020-11-03
Delete account functionality

sroderick-g5sro:
Fixes: #192


sroderick-g5sro:
This code appears to correctly delete an account and files. However the empty directories are left.

weblab/accounts/tests/test_signals.py
```diff
@@ -25,3 +25,20 @@ def test_email_sent_on_user_creation(superuser):
...
+    model = model_with_version
+    user = model.author
+
+    user_directory_repo = user.get_storage_dir('repo')
+
+    assert user_directory_repo.is_dir()
+
+    response = client.post(
+        '/accounts/%d/delete/' % user.pk,
+    )
```
sroderick-g5sro, line 41: @jonc125 For some reason my call here doesn't either delete the user or their folders. In the gui the user is definitely deleted?

sroderick-g5sro:
I have added breakpoints everywhere and none of them are hit when this is called.
```
response = client.post(
        '/accounts/%d/delete' % logged_in_user.pk,
    )
```

I am wondering if the problem is the way I have wired up the template i.e.

```
class UserDeleteView(LoginRequiredMixin, DeleteView):
    """
    Delete a user
    """
    template_name = 'registration/account_confirm_delete.html'
    model = User
```

jonc125:
Hmm, I don't think there's anything fundamentally wrong with that view definition, and the template setting is unlikely to be the issue (especially if as you say it does the right thing when used interactively in the browser). It would be worth having a look at what is returned as the [`response`](https://docs.djangoproject.com/en/1.11/topics/testing/tools/#testing-responses) to get some more understanding of what the view is doing - for instance looking at the `status_code` field, `content` if it's a `20x` response, `url` if it's a `30x` to see where it redirects you, etc.

Remember that if the delete is successful you'll get redirected to the success_url that you set in the view. So actually maybe it would be best if that was `reverse('home')` to take you to the home page, so it's easy to distinguish from the case where you're not logged in and get redirected to the login page?

sroderick-g5sro:
Thanks Jonathan. Good thinking about the home page I have changed that. The call to post is responding with a 301 status with a redirect to /accounts/2/delete/. I am still at a loss to know why the test is not deleting the user.

jonc125:
Odd. I wondered if it was https://stackoverflow.com/questions/1579846/django-returning-http-301 but you do have a trailing slash on your original URL. Maybe it's a security thing https://stackoverflow.com/questions/12784905/django-test-client-gets-301-redirection-when-accessing-url? Though deleting an experiment doesn't seem to need that!

sroderick-g5sro:
I've written a unit test to see if you can delete another user and it appears you can. Do we need to add some logic to prevent this?

weblab/accounts/tests/test_signals.py
```diff
@@ -25,3 +27,25 @@ def test_email_sent_on_user_creation(superuser):
     assert 'Test User' in body
     assert 'test@example.com' in body
     assert 'http://127.0.0.1:8000/admin/accounts/user/{}/change/'.format(user.pk) in body
+
+
+@pytest.mark.django_db
```
jonc125, line 32: Sorry, should have noticed this before, but I think this test would be better in `accounts/tests/test_views.py` as it checks via the view rather than deleting the user directly.

The functionality of the test looks good though!

weblab/accounts/tests/test_signals.py
```diff
@@ -25,3 +27,25 @@ def test_email_sent_on_user_creation(superuser):
...
+def test_delete_user_directory(client, logged_in_user, my_dataset):
```
jonc125, line 33: Perhaps rename this test given we'll want to add some others:
```suggestion
def test_user_can_delete_own_account(client, logged_in_user, my_dataset):
```

Other tests could be:
* `test_delete_user_requires_login` (the original test you had, without the `logged_in_user`)
* `test_user_cannot_delete_other_account`

jonc125 (changes requested):
I agree that we want to prevent a user deleting any account but their own :)

My guess would be that we can do this by making the view inherit from `UserPassesTestMixin` (in `django.contrib.auth.mixins`) rather than `LoginRequiredMixin`. This requires us to implement a `test_func` method which checks whether the current user is allowed to delete the user specified in the URL. I think something like:
```python
    def test_func(self):
        """A user can only delete their own account."""
        return self.get_object() == self.request.user
```

weblab/accounts/tests/test_views.py
```diff
@@ -1,5 +1,62 @@
...
+@pytest.mark.django_db
+def test_delete_user_requires_login(client, other_user):
+    recipes.model.make(author=other_user)
+    user_directory_repo = other_user.get_storage_dir('repo')
+    assert user_directory_repo.is_dir()
+    response = client.post(
+        '/accounts/%d/delete/' % other_user.pk,
+    )
+    assert response.status_code == 302
+    assert User.objects.filter(pk=other_user.pk).exists()
```
jonc125, line 45: Can we add a test `assert '/login' in response.url` here?

weblab/accounts/tests/test_views.py
```diff
@@ -1,5 +1,62 @@
...
+
+    assert user_directory_repo.is_dir()
+    assert user_directory_dataset.is_dir()
+
+    response = client.post(
+        '/accounts/%d/delete/' % logged_in_user.pk,
+    )
+
+    assert response.status_code == 302
+    assert not User.objects.filter(pk=logged_in_user.pk).exists()
```
jonc125, line 30: We could check where this redirect is to; I think it should be `assert response.url == reverse('home')`?

weblab/accounts/tests/test_views.py
```diff
@@ -1,5 +1,62 @@
...
+    recipes.model.make(author=other_user)
+    user_directory_repo = other_user.get_storage_dir('repo')
+
+    assert user_directory_repo.is_dir()
+
+    response = client.post(
+        '/accounts/%d/delete/' % other_user.pk,
+    )
+
+    assert response.status_code == 302
```
jonc125, line 60: Where does this case redirect to?

weblab/accounts/views.py
```diff
@@ -54,13 +54,17 @@ def get_context_data(self, **kwargs):
...
-class UserDeleteView(LoginRequiredMixin, DeleteView):
-    """
-    Delete a user
-    """
-    template_name = 'registration/delete_user.html'
+class UserDeleteView(UserPassesTestMixin, DeleteView):
+    template_name = 'registration/account_confirm_delete.html'
     model = User
 
+    """
```
jonc125, line 61: The docstring should be the first thing within the class, before `template_name` etc.

jonc125 (changes requested):
I think this is basically there now, but one style issue to fix and a suggestion to beef up the testing slightly.

jonc125 (approved):
Good to merge!

PR #319, opened 2020-11-13
274 link dataset to protocol

helenst:
For #274 - provides a tabular form for linking dataset columns to protocol ioputs, grouped by protocol version.


weblab/datasets/views.py
```diff
@@ -370,3 +372,110 @@ def get_context_data(self, **kwargs):
...
+            for version, mappings in groupby(existing_mappings, lambda x: x.protocol_version)
+        }
+
+        # Full list of protocol versions for this user
+        protocol_versions = dataset.protocol.cachedentity.versions.visible_to_user(self.request.user)
+
+        for version in protocol_versions:
+            self.forms[version] = []
+
+            # Each form will ist all but FLAG ioputs for this protocol version
```
jonc125, line 420:
```suggestion
            # Each form will list all except FLAG ioputs for this protocol version
```

weblab/datasets/views.py
```diff
@@ -370,3 +372,110 @@ def get_context_data(self, **kwargs):
...
+            ioputs = version.ioputs.exclude(kind=(ProtocolIoputs.FLAG))
+
+            # Group existing mappings for this version by column name
```
jonc125, line 423: 'Group' might be misleading, since the unique_together means we'll never have multiples?

weblab/datasets/views.py
```diff
@@ -370,3 +372,110 @@ def get_context_data(self, **kwargs):
...
+            for (i, column_name) in enumerate(columns):
+                instance = mappings.get(column_name)
+                form = DatasetColumnMappingForm(
+                    prefix='mapping_%d_%d' % (version.pk, i),
+                    data=self.request.POST or None,
+                    instance=instance,
+                    dataset=dataset,
+                    protocol_ioputs=ioputs,
+                    initial={
+                        'dataset': dataset,
```
jonc125, line 440: The column name isn't being passed in, so always appears blank in the browser!

weblab/datasets/views.py
```diff
@@ -370,3 +372,110 @@ def get_context_data(self, **kwargs):
...
+                    },
+                )
+                self.forms[version].append(form)
+
+        return self.forms
+
+    @property
+    def all_forms(self):
+        """Iterate through all forms in the view"""
+        for _, forms in self.forms.items():
```
jonc125, line 451: Use `.values()` instead?

weblab/datasets/views.py
```diff
@@ -370,3 +372,110 @@ def get_context_data(self, **kwargs):
...
+
+    def get_context_data(self, **kwargs):
+        kwargs['forms'] = self.get_forms()
+        return super().get_context_data(**kwargs)
+
+    def get_success_url(self):
+        """What page to show when the form was processed OK."""
+        dataset = self.object
+        ns = self.request.resolver_match.namespace
+        return reverse(ns + ':map_columns', args=[dataset.id])
```
jonc125, line 472: It might be nice to display a message on the returned page so that it's obvious to the user that their settings have been saved. Otherwise they might think nothing happened?

weblab/datasets/tests/test_models.py
```diff
@@ -37,6 +66,19 @@ def test_visibility_and_sharing(self, logged_in_user, other_user, anon_user):
 
         # TODO - No testing of shared datasets - waiting for implementation in front end
 
+    def test_column_names(self, dataset_creator):
```
jonc125, line 69: It would be worth adding a test with a tab-separated file too, since some of our datasets are (but still called .csv). Perhaps also a test like `b'"column 1","column 2"'`.

And one with two files where the second is the main file?

weblab/datasets/tests/test_views.py
```diff
@@ -768,3 +768,125 @@ def test_ensure_private_results_are_not_shown(self, client, public_dataset):
...
+                prefix + 'dataset': dataset.pk,
+                prefix + 'column_name': 'col',
+                prefix + 'column_units': 'meters',
+                prefix + 'protocol_version': proto_v1.pk,
+                prefix + 'protocol_ioput': proto_v1_in.pk,
+            },
+        )
+
+        assert response.status_code == 302
+        assert dataset.column_mappings.count() == 1
```
jonc125, line 860: Maybe check it fills the correct values in?

jonc125 (changes requested):
This looks good overall. There are a few usability niggles I've spotted that need addressing, and a couple of places where tests could be strengthened.

weblab/datasets/views.py
jonc125: I'm not seeing the message appear. Is the necessary JS hook missing in the datasets templates?

helenst:
OK - I think I've addressed everything here! Also rebased to resolve conflicts.

jonc125 (approved):
All looks good to me here now! Thanks Helen.

Issue #318, opened 2020-11-03
Drop-down nav menus should overlay page text


jonc125:
They get obscured by long entity names, for instance, and by the notifications box. This should just be a matter of tweaking the (S)CSS, particularly the [navbar-submenu class](https://github.com/ModellingWebLab/WebLab/blob/master/weblab/static/sass/style.scss#L138). Using z-index might do it, although apparently it can be quirky to get right! We'd need to add `position: relative` as well I think.

Examples of the problem:

![Screenshot 2020-11-16 at 16 35 30](https://user-images.githubusercontent.com/2032871/99280902-ca6f6400-2829-11eb-9899-09bb47bb8417.png)

![Screenshot 2020-11-16 at 16 34 34](https://user-images.githubusercontent.com/2032871/99280815-ac096880-2829-11eb-8c36-88ac9790062b.png)

PR #322, opened 2020-11-18
Changed sub menu css to stop overlay

sroderick-g5sro:
Fixes: #318

### DIFF
--- a/weblab/static/sass/style.scss
+++ b/weblab/static/sass/style.scss
@@ -139,6 +139,7 @@ button,
   min-width: 100px;
   position: relative;
   z-index: 99;
+  background-color: #ffffff;
 
   ul {
     display: none; /* Hides the drop-down menu */
--- a/weblab/static/sass/style.scss
+++ b/weblab/static/sass/style.scss
@@ -137,6 +137,8 @@ button,
 
 .navbar-submenu {
   min-width: 100px;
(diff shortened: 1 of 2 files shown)


jonc125 (changes requested):
This isn't working for me on either Firefox or Chrome, despite confirming that gulp has re-generated style.css and the browser is loading the new version. I wonder whether the CSS change needs to be applied to the `ul` element inside `.navbar-submenu`?

sroderick-g5sro:
Odd it seems to be working for me as below: 

image

image

jonc125:
Ah, OK. So it's on top of the notifications pane for instance, but because the menu background is transparent it's still hard to read, particularly for the 'on top of entity name' case. Is it possible to tweak the background settings so we get the submenu in a solid plane white box?

sroderick-g5sro:
I have added an white background to each menu but the text "fitting specification" falls outside as shown below: I have tried making the list boxes wider but as the title of each is centred it looks strange and disconnected. So we can either live with the last bit of "specifications" being outside a white background. Or perhaps shorten the text to "fitting specs"?

image

jonc125:
I'd be inclined to go with the shorter text option I think.

jonc125 (approved):
Looks great!

Issue #267, opened 2020-06-29
Versions of experiments


MichaelClerx:
The experiments have versions now, which is a bit weird? If nothing is broken, every version should be the same. What I would much prefer instead of "list versions" is "list protocol versions" and "list model versions", so that I can see how the results compare to previous models/protocols

jonc125:
Versions can easily differ if the underlying software has been updated in any way...

Also sometimes the first run just hit an unexpected error (network timeout or whatever) so had no results.

jonc125:
But adding extra links to the existing views could be done.

jonc125:
@MichaelClerx for instance, we could link to [this matrix](https://scrambler.cs.ox.ac.uk/experiments/models/11/versions/*/protocols/2/versions/*) from a [single experiment view](https://scrambler.cs.ox.ac.uk/experiments/26419/versions/26811/outputs_All_state_variables_gnuplot_data.csv/displayPlotFlot). But what would we call it and where would we put the link? Alongside 'list all versions' but called 'compare results from other model/protocol versions'?

MichaelClerx:
That sounds good! Or even just "see" instead of "compare" !

jonc125:
So I think we'd want a template tag which has similar functionality to `url_version_comparison_matrix` in `experiments.py` but that takes both a model and a protocol as arguments (and so is a `simple_tag` rather than a `filter`). And you'd need to use that to construct a link within `experimentversion_detail.html`.

PR #325, opened 2020-11-20
Added link to comparison matrix

sroderick-g5sro:
Fixes: #267

### DIFF
--- a/weblab/experiments/templatetags/experiments.py
+++ b/weblab/experiments/templatetags/experiments.py
@@ -60,7 +60,7 @@ def url_version_comparison_matrix(entity):
 
 
 @register.simple_tag
-def url_protocol_model_version_comparison_matrix(model, protocol):
+def url_model_protocol_version_comparison_matrix(model, protocol):
     """
         Build URL for model/protocol versions submatrix view
     """
--- a/weblab/experiments/templatetags/experiments.py
+++ b/weblab/experiments/templatetags/experiments.py
@@ -60,13 +60,13 @@ def url_version_comparison_matrix(entity):
 
 
 @register.simple_tag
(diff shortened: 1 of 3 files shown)


weblab/experiments/templatetags/experiments.py
```diff
@@ -59,6 +59,19 @@ def url_version_comparison_matrix(entity):
     return reverse('experiments:list', kwargs=kwargs)
 
 
+@register.simple_tag
+def url_protocol_model_comparison_matrix(experiment):
+    kwargs = {}
+
+    kwargs['model_pks'] = '/%d' % experiment.model.pk
+    kwargs['model_versions'] = '/*'
```
jonc125, line 69: This doesn't quite work because you get URLs looking like `/experiments/models/1/versions/*protocols/42/versions/*`, i.e. with no slash between the `*` and `protocols`. I suspect just setting model_versions to `/*/` would resolve this!

jonc125: But we should also add a test that the templatetag produces expected output given particular arguments, probably in `experiments/tests/test_templatetags.py` `test_url_version_comparison_matrix`

jonc125 (changes requested):

weblab/experiments/templatetags/experiments.py
sroderick-g5sro: good points thanks

jonc125 (approved):
This works nicely now.

One nit-picky change but otherwise good to merge: could we name the tag `url_model_protocol_version_comparison_matrix` rather than `url_protocol_model_version_comparison_matrix` given that the arguments have to be ordered model then protocol?

PR #326, opened 2020-11-25
No sharing (collaborators) functionality for datasets

sroderick-g5sro:
Fixes: #247


sroderick-g5sro:
@jonc125 I have tried to follow the instructions on the issue but have got stuck again - apologies. Please would you mind having a quick look at the following files to see if I am heading in the right direction:

dataset_header.html
datasets\views.py - DatasetCollaboratorsView
datasets\urls.py
templates\datasets\dataset_collaborators_form.html

weblab/datasets/views.py
```diff
@@ -486,3 +487,10 @@ def post(self, request, *args, **kwargs):
             return HttpResponseRedirect(self.get_success_url())
         else:
             return self.render_to_response(self.get_context_data(forms=forms))
+
+
+class DatasetCollaboratorsView(EntityCollaboratorsView):
```
jonc125, line 492: Ah, I've realised an issue with just inheriting here. `EntityCollaboratorsView` inherits from `EntityTypeMixin` which we *don't* want to do. So possibly we need another mixin created in `entities/views.py` that contains the common fields/methods, and which both concrete views can inherit from.

So perhaps a new `EditCollaboratorsAbstractView` class, which contains everything from `EntityCollaboratorsView` except:
* no `template_name` field
* get_context_data is changed to remove the `kwargs['type'] = self.object.entity_type` line

Then we could have
```python
class EntityCollaboratorsView(EntityTypeMixin, EditCollaboratorsAbstractView):
    template_name = 'entities/entity_collaborators_form.html'

    def get_context_data(self, **kwargs):
        kwargs['type'] = self.object.entity_type
```
and
```python
class DatasetCollaboratorsView(EditCollaboratorsAbstractView):
    template_name = 'datasets/dataset_collaborators_form.html'
```

weblab/templates/datasets/includes/dataset_header.html
```diff
@@ -16,7 +16,8 @@ <h1 id="entityname">
         <img src="{% static 'img/delete.png' %}" alt="delete dataset"
              title="delete dataset"/></a>
     {% endif %}
-
+    <a title="manage collaborators for this dataset" href="{% url 'datasets:entity_collaborators' dataset.id %}">
+        Manage collaborators
```
jonc125, line 20: I think we need a `</a>` here, and possibly a full stop!

weblab/datasets/views.py
sroderick-g5sro: thanks. Ive implemented that but I now get the following error:

`DatasetCollaboratorsView is missing a QuerySet. Define DatasetCollaboratorsView.model, DatasetCollaboratorsView.queryset, or override DatasetCollaboratorsView.get_queryset().`

jonc125: Ah yes, that's given by EntityTypeMixin for the other view; we need to set `model = Dataset` as a field for the datasets one.

sroderick-g5sro:
I have now added in `get_context_data` and the form is working. However when I try and assign a dataset to another user I get a permission matching query does not exist error.

jonc125:
Interesting, because using the add_collaborator method directly we’ve tested and it works. It’d definitely be worth creating a failing test that uses the new view to add a collaborator so we have something more straightforward to debug. Possibly going via the form (which lives in entities) has caused Django to get confused about which app it’s looking for the edit_entity permission in?

sroderick-g5sro:
apologes @jonc125 I have attempted to write a unit test going via the form as below (so we can debug the permissions issue):

  ```
  def test_can_share_dataset(self, logged_in_user, other_user, public_protocol, client):
        shared_dataset = recipes.dataset.make(
            author=logged_in_user, name='mydataset', visibility='private', protocol=public_protocol)
        response = client.post('/datasets/%d/collaborators' % shared_dataset.pk,
                               {
                                   'form-0-email': other_user.email,
                                   'form-TOTAL_FORMS': 1,
                                   'form-MAX_NUM_FORMS': 1000,
                                   'form-MIN_NUM_FORMS': 0,
                                   'form-INITIAL_FORMS': 0,
                               })
```
but I get the following error message.

```
def get_success_url(self):
        """What page to show when the form was processed OK."""
        entity = self.object
        ns = self.request.resolver_match.namespace
>       return reverse(ns + ':entity_collaborators', args=[entity.url_type, entity.id])
E       AttributeError: 'Dataset' object has no attribute 'url_type'
```

jonc125:
Ah, so I missed one point of difference between entity & dataset with the success URL, which means we will need this method to be defined different in each view. In the dataset version you don't need the `entity.url_type` argument, just the dataset ID is sufficient to define the URL. And you could call the local variable `dataset` rather than `entity` in that case for clarity!

sroderick-g5sro:
> Interesting, because using the add_collaborator method directly we’ve tested and it works. It’d definitely be worth creating a failing test that uses the new view to add a collaborator so we have something more straightforward to debug. Possibly going via the form (which lives in entities) has caused Django to get confused about which app it’s looking for the edit_entity permission in?

The unit test that I have created seems to work as expected but I still get permissions error occuring when I run it in the browser.

jonc125:
Annoyingly Travis has stopped running our tests due to their recent policy change :(  But running locally while there are 6 failing dataset tests, your new one is passing. When do you get the permissions error - when you view the form or when you submit it?

sroderick-g5sro:
I get the error on submit

jonc125:
The problem was missing DB migrations adding the permission in the datasets app. It works for tests because they create their own DB, and so don't need to run migrations. But testing live on the dev server does need the DB updated.

sroderick-g5sro:
Great thanks

jonc125:
@sroderick-g5sro could you resolve the merge conflicts, and I'll review after that?

jonc125:
I can confirm the tests pass locally. (I've also requested more build minutes from Travis CI!)

sroderick-g5sro:
Ok hopefully I have got the merge correct. I have run the tests locally and they all pass.

jonc125 (approved):
Works well here too!

Issue #315, opened 2020-11-01
Remove hacky fitting spec implementation from front-end


jonc125:
Once #314 merged and deployed, and key fitting results from old implementation replicated in the new one.

@mirams is it just the single Kylie sine wave fit to cell 5 we care about? Or are there others I should run?

mirams:
Once that's working I think it's OK, when the fitting spec is up and running we can try that with other datasets. Or are there specific things you'd like to test - like different length datasets or different protocols or anything? We can easily make some synthetic data to try other cases even if we don't have real stuff

PR #327, opened 2020-11-25
Remove references to is_fitting_spec

sroderick-g5sro:
Fixes: #315


weblab/entities/migrations/0016_delete_fitting_spec_protocols.py
```diff
@@ -0,0 +1,21 @@
...
+
+from django.db import migrations
+
+def delete_protocols(apps, schema_editor):
+
+    Protocol = apps.get_model('entities', 'ProtocolEntity')
+    for protocol in Protocol.objects.all():
+        if protocol.is_fitting_spec:
+            protocol.delete();
+
```
sroderick-g5sro, line 11: I have not tested this yet also do I need to put in the reverse migration

jonc125: We don't need the reverse migration - there's not really anything we can do for it given we delete data!

One suggestion on the above: you can use `Protocol.objects.filter(is_fitting_spec=True)` to do the filtering in the DB rather than in Python.

weblab/entities/migrations/0016_delete_fitting_spec_protocols.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 2.2.16 on 2020-11-25 15:02
+
+from django.db import migrations
+
+def delete_protocols(apps, schema_editor):
```
jonc125, line 5: I'd call this `delete_old_style_fitting_specs` to make it clearer what it's doing.

weblab/templates/entities/includes/entity_header.html
```diff
@@ -2,7 +2,7 @@
 {% load entities %}
 
 <h1 id="entityname">
-  <a href="{% ns_url 'detail' entity.url_type entity.id %}"><small>{% if entity.is_fitting_spec %}Parameter fitting specification{% else %}{{ type|capfirst }}{% endif %}: </small><span>{{ entity.name }}</span></a>
+  <a href="{% ns_url 'detail' entity.url_type entity.id %}"><span>{{ entity.name }}</span></a>
```
jonc125, line 5: We need to keep the content of the `else` clause of the `if` here.

jonc125 (commented):
Not had a chance to try this out or look at the tests yet, but it looks along the right lines!

weblab/entities/migrations/0016_delete_fitting_spec_protocols.py
sroderick-g5sro: thanks Jonathan. I have made those changes. Do I need to change/write any unit tests?

weblab/entities/migrations/0016_delete_old_style_fitting_specs.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.2.16 on 2020-11-25 15:02
+
+from django.db import migrations
+
+
+def delete_protocols(apps):
```
jonc125, line 6: It was this function I meant to rename, rather than the whole file. (Not that renaming the file is a problem!)

weblab/entities/migrations/0016_delete_old_style_fitting_specs.py
```diff
@@ -0,0 +1,19 @@
...
+    Protocol = apps.get_model('entities', 'ProtocolEntity')
+    for protocol in Protocol.objects.filter(is_fitting_spec=True):
+        protocol.delete();
```
jonc125, line 9: We don't need ; in Python :)

jonc125 (changes requested):
I don't think we need to update any tests or write new ones, given everything is passing.

I've discovered a bug running the migrations on my local dev setup though:
```
TypeError: delete_protocols() takes 1 positional argument but 2 were given
```
You seem to be missing the `schema_editor` argument in your function definition.

weblab/entities/migrations/0016_delete_old_style_fitting_specs.py
sroderick-g5sro: apologies... I tried to retest but not sure how to "rollback" migrations. when I try and specify an earlier migration I get error message 

`django.db.migrations.exceptions.InconsistentMigrationHistory: Migration entities.0017_remove_entity_is_fitting_spec is applied before its dependency entities.0016_delete_old_
style_fitting_specs on database 'default'.`

jonc125: Because we didn't provide a reverse migration for 0016 it isn't reversible, so you can't go back to before that state. I'll test on a VM here, but looks like it should be OK now.

jonc125 (approved):
Migration ran successfully, and it appears to be behaving well.

Issue #328, opened 2020-11-26
Allow dataset visibility to be changed after creation


jonc125:
With models/protocols/fitting specs, you get a 'Change visibility: ' + drop-down option for entities you have edit rights for, just after the 'Add tag' link. For datasets there's just static text showing the current visibility as part of the header.

1. We should replace this static text with a change interface if the logged in user has permissions to do so.
2. Also, since datasets don't have versions, the header (created...by...) info is duplicated. It should only appear under the main title, not inside the box.
3. And the visibility section should probably come on the same line as the 'corresponding protocol' bit, at the start of the line.

The latter two just require changing `templates/datasets/base.html`.

The first requires creating a version of `entities.views.ChangeVisibilityView` for datasets. It can use the same `EntityChangeVisibilityForm` form, but will need a different `post` method because datasets don't have `set_version_visibility` - the visibility is directly on the dataset. `DatasetView` will need to gain a `get_context_data` method similar to `EntityVisibilityFormMixin` (but again with getting visibility direct from the dataset). And `templates/datasets/base.html` will need to gain the change visibility content from `templates/entities/entity_version.html`.

All these changes will need appropriate tests.

PR #329, opened 2020-12-02
Allow dataset visibility to be changed after creation

sroderick-g5sro:
Fixes: #328


jonc125 (approved):
This is working nicely for me. I tweaked the display slightly - decided it was easier just to do it than explain what I wanted! Only one very minor comment.

PR #330, opened 2020-12-04
Added pipe separator between headings

sroderick-g5sro:
Fixes part of issue #239


weblab/templates/datasets/includes/dataset_header.html
```diff
@@ -11,14 +11,15 @@ <h1 id="entityname">
   <small>Created <time>{{dataset.created_at}}</time> by <em>{{dataset.author.full_name}}</em>.
     {% can_delete_entity dataset as can_delete %}
     {% if can_delete %}
-      Delete dataset:
+
       <a title="delete dataset" href="{% url 'datasets:delete' dataset.id %}">
+          Delete dataset:
```
jonc125, line 16: Now these are part of the link I'd remove the colon.

weblab/templates/datasets/includes/dataset_header.html
```diff
@@ -11,14 +11,15 @@ <h1 id="entityname">
...
         <img src="{% static 'img/delete.png' %}" alt="delete dataset"
              title="delete dataset"/></a>
     {% endif %}
 
     <a title="Transfer ownership of dataset" href="{% url 'datasets:transfer' dataset.id %}">
-      Transfer ownership </a>
+    | Transfer ownership |</a>
```
jonc125, line 22: I'd display the first vertical bar within the if can_delete, otherwise it'll look a bit odd for users who can't delete the dataset.

Though it occurs to me the transfer and rename links also need to be within an if block, since only the owner should be allowed to do the first, and only someone with edit rights should be able to do the second. Or possibly anyone able to manage the dataset can do both, as with models/protocols in the file below?

weblab/templates/entities/includes/entity_header.html
```diff
@@ -9,19 +9,19 @@ <h1 id="entityname">
   <small>Created <time>{{entity.created_at}}</time> by <em>{{entity.author.full_name}}</em>.
     {% can_delete_entity entity as can_delete %}
     {% if can_delete %}
-      Delete {{ type }}:
       <a title="delete all versions of this {{ type }}" href="{% entity_url 'delete' entity %}">
+      Delete {{ type }}:
```
jonc125, line 13:
```suggestion
      Delete {{ type }}
```

weblab/templates/entities/includes/entity_header.html
```diff
@@ -9,19 +9,19 @@ <h1 id="entityname">
...
         <img src="{% static 'img/delete.png' %}" alt="delete all versions of this {{ type }}"
              title="delete all versions of this {{type}}"/></a>
     {% endif %}
     {% can_manage_entity entity as can_manage %}
     {% if can_manage %}
       &nbsp;
       <a title="manage collaborators for this {{ type }}" href="{% entity_url 'entity_collaborators' entity %}">
-        Manage collaborators
+      |  Manage collaborators |
```
jonc125, line 21: Again, move the first vertical bar into the delete block.

weblab/templates/entities/includes/entity_header.html
```diff
@@ -9,19 +9,19 @@ <h1 id="entityname">
...
       </a>
       <a title="Transfer ownership of {{ type }}" href="{% entity_url 'transfer' entity %}">
-        Transfer ownership
+        Transfer ownership |
       </a>
 
     {% endif %}
```
jonc125, line 27: The rename link below should also move into this if.

weblab/templates/datasets/includes/dataset_header.html
```diff
@@ -11,14 +11,15 @@ <h1 id="entityname">
...
     <a title="Rename dataset" href="{% url 'datasets:rename' dataset.id %}">
         Rename dataset
       </a>
```
jonc125, line 25: Looking at the display of datasets compared with fitting specs, the latter has the linked protocol displayed at the entity header level (and called 'Linked protocol'), whereas datasets have it displayed inside the box (and called 'Corresponding protocol'). We should probably be consistent, and I think change the datasets to match fitting specs is best (since we shouldn't change where it sits for specs as it's a property of the whole spec not a single version).

jonc125 (changes requested):
This starts to look at the right areas, but I think we can go further to improve consistency and clarity, particularly for users who aren't logged in yet. I've made various specific comments. More broadly:
* These links use an initial capital letter, which I like. Various links elsewhere (e.g. the 'list all versions' links, 'export plot data', 'zoom', 'select all', 'compare selected' etc use all lower case. Unless you and/or @mirams object, I think I'd say change all links to have an initial capital letter? There will be others I've not mentioned, so have a browse around the live site (https://chaste.cs.ox.ac.uk/WebLab) to see what you can find that looks mismatched.
    * Similarly we have 'Created' here, but 'created' in the version lists for entities and experiments.
* In the entity version view, we should probably change the 'Run experiments' text to be a link like for 'Delete'. And I wonder about moving the 'Add tag' link next to that one, rather than at the start of the row?
* 'Delete experiment' and 'Delete experiment version' should also link-ify the text. And perhaps we should add 'Re-run' text alongside the icon for that action?

sroderick-g5sro:
Hopefully I have got all the required changes.

jonc125 (approved):

PR #349, opened 2021-06-16
fixes JS error when running in subfolder

MauriceHendrix:


weblab/core/processing.py
```diff
@@ -0,0 +1,11 @@
+from urllib.parse import urljoin
+
+from django.conf import settings
+
+def prepend_callback_base(url_path):
+    """Prepend CALLBACK_BASE_UR to url_path. If we are running in a subfolder this is striped from url_path."""
+
+    if hasattr(settings, 'FORCE_SCRIPT_NAME'):
+         url_path = url_path.replace(settings.FORCE_SCRIPT_NAME, '')
```
jonc125, line 9: Just in case someone chooses a prefix that also appears later in some URLs, I think I'd do this instead:
```suggestion
    if hasattr(settings, 'FORCE_SCRIPT_NAME') and url_path.startswith(settings.FORCE_SCRIPT_NAME):
         url_path = url_path[len(settings.FORCE_SCRIPT_NAME):]
```

weblab/static/js/db.js
```diff
@@ -20,7 +20,8 @@ function submitNewExperiment(jsonObject, $td, entry)
 
   $td.append("<img src='"+staticPath+"img/loading2-new.gif' alt='loading' />");
 
-  $.post('/experiments/new', jsonObject, function(data) {
+//  $.post('/experiments/new', jsonObject, function(data) {
+  $.post($("#matrixdiv").data('new-href'), jsonObject, function(data) {
```
jonc125, line 24:
```suggestion
  $.post($("#matrixdiv").data('new-href'), jsonObject, function(data) {
```

Similarly elsewhere - don't keep commented out lines :)

weblab/core/processing.py
```diff
@@ -0,0 +1,11 @@
+from urllib.parse import urljoin
+
+from django.conf import settings
+
+def prepend_callback_base(url_path):
+    """Prepend CALLBACK_BASE_UR to url_path. If we are running in a subfolder this is striped from url_path."""
```
jonc125, line 6: How about this?
```suggestion
    """Prepend CALLBACK_BASE_URL to url_path.
    
    If we are running at a sub-URL (settings.FORCE_SCRIPT_NAME is set) then the prefix is stripped,
    since the callback is done on a private interface (to localhost) not via the public WebLab address.
    """
```

weblab/core/processing.py
MauriceHendrix: I think this should od it too?:
```
url_path = url_path.replace(settings.FORCE_SCRIPT_NAME, '', `)
```

README.md
```diff
@@ -91,9 +91,15 @@ The Ansible deployment will install the XML/RDF file from https://github.com/Mod
...
 
-## Run unit tests
+## Run unit tests (needs requirements installed e.g. in virtual environment)
 
 ```
 cd weblab
 pytest
 ```
+
+If you encounter database access issues. You may need to give the weblab user rights to create databases:
```
jonc125, line 101:
```suggestion
If you encounter database access issues, you may need to give the weblab user rights to create databases:
```

weblab/config/settings/test.py
```diff
@@ -1,4 +1,5 @@
 from .base import *  # noqa
+from .deployed import DATABASES
```
jonc125, line 2: Don't commit this - it's liable to break the CI tests. It should only be included on Vagrant VM deployments, which should be done in the Ansible configuration instead.

weblab/entities/processing.py
```diff
@@ -5,6 +5,7 @@
 from django.conf import settings
 from django.db import IntegrityError
 from django.urls import reverse
+from core.processing import prepend_callback_base
```
jonc125, line 8: The blank line should come before this import, not afterwards - it's part of our codebase not Django's.

weblab/experiments/processing.py
```diff
@@ -7,6 +7,7 @@
 from django.core.exceptions import MultipleObjectsReturned
 from django.urls import reverse
 from django.utils.timezone import now
+from core.processing import prepend_callback_base
```
jonc125, line 10: Same here.

weblab/fitting/processing.py
```diff
@@ -4,6 +4,8 @@
 from django.conf import settings
 from django.core.exceptions import MultipleObjectsReturned
 from django.urls import reverse
+from core.processing import prepend_callback_base
```
jonc125, line 7: And here.

weblab/core/processing.py
```diff
@@ -0,0 +1,14 @@
+from urllib.parse import urljoin
+
+from django.conf import settings
+
+def prepend_callback_base(url_path):
```
jonc125, line 5: I recall you were going to write a test for this; is that still todo or just forgot to add the file?

MauriceHendrix: the test is here: `core/tests/test_processing.py`

weblab/core/tests/test_processing.py
```diff
@@ -0,0 +1,23 @@
...
+from django.conf import settings
+from core.processing import prepend_callback_base
+
+
+@pytest.mark.django_db
+def test_prepend_callback_base():
+    old_callback_base = settings.CALLBACK_BASE_URL
+    old_script_name = settings.FORCE_SCRIPT_NAME
+    settings.CALLBACK_BASE_URL = 'http://test.com/'
+    settings.FORCE_SCRIPT_NAME = None
```
jonc125, line 12: It's better to use the settings fixture for this, as it will automatically restore the old settings however the test exits. See https://pytest-django.readthedocs.io/en/latest/helpers.html#settings

jonc125 (changes requested):
Looks good generally; just suggested a better way of handling settings in the test.

weblab/config/settings/test.py
MauriceHendrix: or I can just do
```
try:
    from .deployed import DATABASES  # noqa
except ImportError:
    pass
```
a bit crude but it does work

MauriceHendrix: added vagrant.py with
```
from .base import *  # noqa
from .deployed import DATABASES  # noqa
```
and added instructions

jonc125 (approved):

PR #350, opened 2021-06-18
Create pytest.yml

MauriceHendrix:
moved to using github actions. 
Please note: it doesn't work for python 3.9 currently as the psycopg2 version isn't compatibale with python 3.9 but the version that is has a known bug in it :(

edit: it also seems to not work for python 3.8


MauriceHendrix:
moved to using github actions. 
Please note: it doesn't work for python 3.9 currently as the psycopg2 version isn't compatibale with python 3.9 but the version that is has a known bug in it :(

weblab/core/tests/test_combine.py
```diff
@@ -87,7 +87,8 @@ def test_xml_doc(self, writer, manifest):
         writer.add_file('test.txt', fmt='txt')
 
         doc = ET.tostring(writer.xml_doc.getroot()).decode()
-        assert doc == manifest
+        manifest_str = ET.tostring(ET.fromstring(manifest)).decode()
+        assert doc == manifest_str is not None
```
jonc125, line 91: This seems a very odd test! A comment explaining why would be helpful!

MauriceHendrix: It's because the order of properties is different in python 3.8 and I'm trying to find a way that can deal with that. This seems not to be working though.

jonc125: Oh fun :(
I'd stick to testing on just 3.5-3.7 for now, and open another issue for supporting 3.8 & 3.9. We control deployments so can control what version is used much more easily.

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,51 @@
...
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip wheel
+        pip install flake8 pytest pytest-django
+        pip install -r requirements/test.txt
+    - name: Lint with flake8
+      run: |
+        python -m flake8 weblab
+    - name: Test sorting wioth isort
```
jonc125, line 38:
```suggestion
    - name: Test sorting with isort
```

jonc125 (approved):

PR #353, opened 2021-06-22
Combined dependabot JS package updates

jonc125:
These don't appear to cause any issues on my local vagrant install...


MauriceHendrix (approved):
Thanks seems to work fine! (I tried it on cardiac.nottingham)

PR #356, opened 2021-06-23
Resolve race condition in loading annotator

### LINKED ISSUE #338: Metadata annotation editor isn't showing existing tags as shaded any more
The metadata annotation tool used to show shaded things for any already-tagged metadata. It doesn't any more, e.g. [Luo-Rudy 1991](https://scrambler.cs.ox.ac.uk/entities/models/3/versions/785d581c1eb54f0bba52daeeaa10adeefe94c479/luo_rudy_1991.cellml/editMetadata)

which looks like this:

image

when you add some new metadata it shades in the right hand entries...

jonc125:
The used ontology terms should now always appear shaded. Fixes #338 - tested locally.


MauriceHendrix (approved):
I haven't tested this but it certainly looks plausable and if it works locally ...

mirams (approved):

PR #358, opened 2021-06-23
Try updating all Python dependencies

jonc125:


MauriceHendrix:
it doesn't even seem to register the python 3.5 task, very odd.
It might be worth considering if we should drop python 3.5 support anyway as a lot of the newest packages have dropped support for it so it's going to be harder nd harder to make exceptions for it?

jonc125:
This PR does drop Python 3.5 support, which is why it hasn't built. I guess it's still showing up because the main branch has it though?

MauriceHendrix:
> This PR does drop Python 3.5 support, which is why it hasn't built. I guess it's still showing up because the main branch has it though?

aah fair enough

MauriceHendrix (approved):

PR #361, opened 2021-07-20
add author name to entity list view

MauriceHendrix:
I added teh author name in brackets to the entity list view, because models at least can have duplicate names, as long as the authors are different.
Here is an example fo what it'll look like:
image

### DIFF
--- a/weblab/templates/entities/entity_list.html
+++ b/weblab/templates/entities/entity_list.html
@@ -23,7 +23,7 @@
     <ul>
       {% for entity in object_list %}
         <li title="{{ entity.name }}">
-          <strong><a href="{% entity_url 'detail' entity %}">{{ entity.name }}</a></strong>
+          <strong><a href="{% entity_url 'detail' entity %}">{{ entity.name }}</strong><em style="margin-right:0.25em"> (by {{ entity.author.full_name }}) </em></a>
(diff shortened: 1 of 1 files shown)


jonc125 (changes requested):
We definitely don't want custom CSS within the templates - it should go in the sass files if needed. Though I'm not sure why it is here.

Looking at the screenshot, it's rather odd having a section titled 'Your models' with an author listed for each! We could (a) retitle it 'Models you can see' or (b) split into two sections 'Your models' and 'Models shared with you'. Which seems better @MichaelClerx @mirams ? I'd suggest also omitting the author name for models you own, and only displaying when the author is someone else.

MauriceHendrix:
> We definitely don't want custom CSS within the templates - it should go in the sass files if needed. Though I'm not sure why it is here.
> 
> Looking at the screenshot, it's rather odd having a section titled 'Your models' with an author listed for each! We could (a) retitle it 'Models you can see' or (b) split into two sections 'Your models' and 'Models shared with you'. Which seems better @MichaelClerx @mirams ? I'd suggest also omitting the author name for models you own, and only displaying when the author is someone else.

pw cr.. you're right I do think it is only your models for this view. Forget about this!

PR #362, opened 2021-07-23
User stories

MauriceHendrix:
Right so I think Model Groups are mostly done (though needs better tests still). Attached you see:
1. overview of model groups
2. creation form (get there by clicking create in overview)
3. edit form (get there by clicking on model in overview)
4. overview for a different author without create privileges but with some shared model groups
5. edit for for a shamed model for this author: visibility is disabled and models they don't have viewing rights for are disabled.
The reason I decided to include model groups that you collaborate on on the overview is that I wanted a way for collaborators to edit them.
Let me know your thoughts,


MichaelClerx:
How did this PR get to 1655 commits?

weblab/entities/forms.py
```diff
@@ -180,3 +180,51 @@ class FileUploadForm(forms.ModelForm):
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Only show models I can can see
+        self.current_title = None
+        self.fields['models'].queryset = ModelEntity.objects.visible_to_user(self.user)
+
+        # Save current details if we have them
+        if 'instance' in kwargs and kwargs['instance']:
```
MichaelClerx, line 200: Worth adding a variable here?
```
instance = kwargs.get('instance', None)
if instance:
    ...
```
?

weblab/entities/forms.py
```diff
@@ -180,3 +180,51 @@ class FileUploadForm(forms.ModelForm):
...
+                visibility.Visibility.MODERATED, 'Moderated')
+            )
+
+    class Meta:
+        model = ModelGroup
+        fields = ['title', 'visibility', 'models']
+
+    def clean_title(self):
+        title = self.cleaned_data['title']
+        if title != self.current_title and self._meta.model.objects.filter(title=title, author=self.user).exists():
```
MichaelClerx, line 219: Do we need to disallow this? Probably doesn't make sense for a user to make groups with same name, but we can still have e.g. groups with the same name but a different owner, or groups with the same name but different capitalisation, or extra spaces etc. So maybe no need to restrict this one?

weblab/entities/models.py
```diff
@@ -453,3 +454,32 @@ class AnalysisTask(models.Model):
...
+    class Meta:
+        ordering = ['title']
+        unique_together = ['title', 'author']
+        permissions = (
+            # edit_modelgroup is used as an object-level permission for the collaborator functionality
+            ('edit_modelgroup', 'Can edit modelgroup'),
+        )
+
+    def visible_to_user(self, user):
+        return self. is_editable_by(user) or user.is_superuser or self.visibility != 'private'
```
MichaelClerx, line 476:
```suggestion
        return self.is_editable_by(user) or user.is_superuser or self.visibility != 'private'
```

MichaelClerx:
Screenshots look good to me except for the checkboxes: I'd expect us to have lots and lots of models, so that you'd want one of those drop-down boxes that you can type in instead?

weblab/entities/forms.py
MauriceHendrix: > ...  but we can still have e.g. groups with the same name but a different owner ...
That's what this does. It checks that if you'ved changed the title it's not the same as one of you own models

MauriceHendrix:
> How did this PR get to 1655 commits?
I was committing as root, when I updated the username on that it seemed to think existing commits wer new ones for some reason.

> Screenshots look good to me except for the checkboxes: I'd expect us to have lots and lots of models, so that you'd want one of those drop-down boxes that you can type in instead?

arggh that would be much easier and is the default but @mirams didn't like it and wanted checkboxes

mirams:
dropdown you can type in to search is fine (as well as checkboxes, it could auto-select the checkboxes?). I just don't like the django-style list of all of them in a scroll window thing that you have to ctrl-click and look down

MauriceHendrix:
> dropdown you can type in to search is fine (as well as checkboxes, it could auto-select the checkboxes?). I just don't like the django-style list of all of them in a scroll window thing that you have to ctrl-click and look down

not quite sure what you mean with auto-select the checkboxes? When editig it obviously pre-selects the ones that are included with the modelgroup if that's what you mean

mirams:
Maurice and I had a chat and decided something like this might be optimal:
image

mirams:
(The above can be a new ticket for improvements later on, doesn't need to slow down merging in if we want to do more user testing)

MauriceHendrix:
see [https://cardiac.nottingham.ac.uk/weblab](https://cardiac.nottingham.ac.uk/weblab) to see it in action

MichaelClerx:
In the first form when you create a story: Graphvisualizer --> Graph visualizer

MichaelClerx:
Select model (group) --> Select model or model group

MichaelClerx:
To render the story, I'm having to save, then use the back button to see the results again.

Can we change the "Save" button to "Save and re-run" + "Save and exit" ?

MauriceHendrix:
> To render the story, I'm having to save, then use the back button to see the results again.
> 
> Can we change the "Save" button to "Save and re-run" + "Save and exit" ?

What do you mean re-run? Do you mean save & go to the render? Would you prefer to always do that (which would be a super easy fix) or only in sime circomstances or...? 
Of course we can have save & view as well as save & exit if you want.

MichaelClerx:
I just want to see what I made! So some button that makes the graphs show up, without sending me to a different page

MichaelClerx:
Can we select the "default graph" by default? Or indicate it somehow in the list? Quite hard to pick the correct one from a list of 10 unfamiliar names

MauriceHendrix:
> I just want to see what I made! So some button that makes the graphs show up, without sending me to a different page



> Can we select the "default graph" by default? Or indicate it somehow in the list? Quite hard to pick the correct one from a list of 10 unfamiliar names

The default graph? How do I know which one is the default one. I've not seen that info anywhere.

MichaelClerx:
@jonc125 how does the normal interface know which graph to show for a particular experiment?

MichaelClerx:
> What do you mean re-run? Do you mean save & go to the render? Would you prefer to always do that also when you first create the story (which would be a super easy fix) or only in sime circomstances or...?
> Of course we can have save & view as well as save & exit if you want.

In jupyter, you can render each cell (text or code) directly after creating. Do you think that's possible? If not, can we do it for graphs at least? I need to be able to see if I picked the right graph after I add one to the story

MauriceHendrix:
> > What do you mean re-run? Do you mean save & go to the render? Would you prefer to always do that also when you first create the story (which would be a super easy fix) or only in sime circomstances or...?
> > Of course we can have save & view as well as save & exit if you want.
> 
> In jupyter, you can render each cell (text or code) directly after creating. Do you think that's possible? If not, can we do it for graphs at least? I need to be able to see if I picked the right graph after I add one to the story

I think it should be probably possible, not quite sure how to style it clearly as a preview but ok.

MichaelClerx (changes requested):
Looks great overal, but two minor features I'd like added to this PR:

- [x] Some way of knowing which CSV set is the "default" or "main" graph in an experiment
- [x] Some way of rendering the graphs (and/or markdown) without leaving the story editor

jonc125:
> @jonc125 how does the normal interface know which graph to show for a particular experiment?

It's the first one listed in outputs-default-plots.csv

MauriceHendrix:
Right so for previews I can come up with 2 ways of showing it. Which one would you prefer @MichaelClerx @mirams @jonc125 @agrawal-aditi .  Either just a  preview icon with a pop-up of the chosen graph (whcn clicked) or a (live) preview panel. With teh panel, I'm a bit worried that it'll tempt authors into spending time editing stuff that doesn't save (zoom level, selected items in teh legent). And it may be a bit bulky. (and I suspect the preview button would be slightly easier to implement.)

image
image

By the way there is a preview already for text:
image

MichaelClerx:
button is fine by me!

MauriceHendrix:
Right graph preview should be working:
image

MichaelClerx (approved):
Thanks Maurice!

PR #367, opened 2021-10-08
redirect to correct page after registration

MauriceHendrix:
The redirection after logging in wasn't going to the correct place, this fixes that

### DIFF
--- a/weblab/accounts/views.py
+++ b/weblab/accounts/views.py
@@ -6,12 +6,13 @@ from django.views.generic.edit import DeleteView, FormView, UpdateView
 
 from .forms import MyAccountForm, RegistrationForm
 from .models import User
+from django.conf import settings
 
 
 class RegistrationView(FormView):
     form_class = RegistrationForm
     template_name = 'registration/register.html'
-    success_url = '/'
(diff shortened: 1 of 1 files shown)


weblab/accounts/views.py
```diff
@@ -6,12 +6,13 @@
 
 from .forms import MyAccountForm, RegistrationForm
 from .models import User
+from django.conf import settings
```
jonc125, line 9: This import should move up a few lines.

weblab/accounts/views.py
```diff
@@ -6,12 +6,13 @@
...
 
 
 class RegistrationView(FormView):
     form_class = RegistrationForm
     template_name = 'registration/register.html'
-    success_url = '/'
+    success_url = settings.FORCE_SCRIPT_NAME if hasattr(settings, 'FORCE_SCRIPT_NAME') else '/'
```
jonc125, line 15: Maybe better:
```suggestion
    success_url = getattr(settings, 'FORCE_SCRIPT_NAME', '/')
```

jonc125 (approved):

PR #374, opened 2021-11-08
Model group functionality only

MauriceHendrix:


MichaelClerx:
What's the difference between this one and the other one?

MauriceHendrix:
> What's the difference between this one and the other one?

this one ONLY has model grouping functionality. If this one gets reviewed first, then merged the other one will be less work.
However if you'd rather review a much bigger PR that's also fine (in that case just review the other one).

MichaelClerx:
OK. I really don't have any way to review this code. I can look at the interface with the link you sent and say if that looks good or not?

MauriceHendrix:
> OK. I really don't have any way to review this code. I can look at the interface with the link you sent and say if that looks good or not?

That would certainly help.

PR #376, opened 2021-11-08
added message about unexpected file type to uploader for dataset

### LINKED ISSUE #375: Better error message when uploading worng type of files for dataset
When I uploaded a cellml file instead of a csv file I got these messages (scrambler and dev box respectively). Probably worth ctaching that error and presenting a better error message.

image
image

MauriceHendrix:
added message about unexpected file type to uploader for dataset
fixes #375 
image

### DIFF
--- a/weblab/templates/datasets/includes/upload.html
+++ b/weblab/templates/datasets/includes/upload.html
@@ -8,7 +8,8 @@
     Drop Files Here<br /> <a>Open Dialog</a>
     <input type="file" name="upload" id="fileupload"
         data-url="{% url 'datasets:upload_file' dataset.id %}"
-        data-form-data='{"csrfmiddlewaretoken": "{{ csrf_token }}"}'>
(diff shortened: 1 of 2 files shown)


weblab/templates/datasets/includes/upload.html
```diff
@@ -8,7 +8,8 @@
     Drop Files Here<br /> <a>Open Dialog</a>
     <input type="file" name="upload" id="fileupload"
         data-url="{% url 'datasets:upload_file' dataset.id %}"
-        data-form-data='{"csrfmiddlewaretoken": "{{ csrf_token }}"}'>
+        data-form-data='{"csrfmiddlewaretoken": "{{ csrf_token }}"}'
+        required-file-type=".csv">
```
jonc125, line 12: I think this should be `data-required-file-type` to confirm with HTML spec/typical usage? And then use `.data` rather than `.attr` to retrieve the value?

PR #377, opened 2021-11-08
add javascript markdown editor

MauriceHendrix:
This PR just installs the SimpleMDE markdown editor, I thought I'd pull this out in a seperate PR as most files probably do not need reviewing. The only thing I changed slightly is simplemde.js (see comments on top as to what's changed)


MauriceHendrix:
> I'm missing here anything about how this is incorporated into the `gulp` build for use by HTML pages. Also, is there a reason `npm` can't be used to install this dependency, or libraries it in turn depends on?

I forgot to include the require statement (now included) Of course it's not used here, but that stuff is in the other PR.

Including source instead of installing via NPM, mostly because I wanted to change the way the library works with images. The default is that if you highlight a text say `my image` then it replaces that by `` which is clearly inadequate. However otherwise it was the closest I could find, to what I wanted.

MichaelClerx:
Fine by me as long as it's OK with @jonc125 .
Do we need to include any license info for these packages on the site somewhere?

MichaelClerx (approved):
Fine by me as long as it's OK with @jonc125 .
Do we need to include any license info for these packages on the site somewhere?

MauriceHendrix:
> Fine by me as long as it's OK with @jonc125 . Do we need to include any license info for these packages on the site somewhere?

good point the license does say that it needs to be included in modified or partial copies, so I included it inside the main js file.

PR #378, opened 2021-12-07
account deletion statement in privacy

MauriceHendrix:
added confirm delete template and updated the mention of deleting accounts in the privacy page

### DIFF
--- a/weblab/templates/registration/account_confirm_delete.html
+++ b/weblab/templates/registration/account_confirm_delete.html
@@ -7,7 +7,7 @@
 
 <form action="" method="post">{% csrf_token %}
     <p>Are you sure you want to delete your account?</p>
-    <p><strong>This operation cannot be undone.</strong></p>
+    <p><strong>This operation cannot be undone. Your fills will be deleted.</strong></p>
     <button>Confirm</button>
     <a class="button" href="javascript:history.back()" title="Do not delete">Cancel</a>
 </form>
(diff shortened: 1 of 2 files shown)


jonc125:
It's probably worth adding to this page a note about what will happen to all their models etc - I'm assuming they get deleted? Do we have a test for that?

jonc125:
Ah, looks like we do have some in `accounts/tests/test_views.py`.

*Not* deleting things the person owns is quite tricky. We'd need to implement a mechanism to transfer ownership first, then give the option to auto-transfer everything when deleting the account.

MauriceHendrix:
> Ah, looks like we do have some in `accounts/tests/test_views.py`.
> 
> _Not_ deleting things the person owns is quite tricky. We'd need to implement a mechanism to transfer ownership first, then give the option to auto-transfer everything when deleting the account.

I just tried it, yes it just gets deleted and actually posentially leeds to an error in stories, but ok that's fixable.

weblab/templates/registration/account_confirm_delete.html
```diff
@@ -0,0 +1,15 @@
+
+{% extends "base.html" %}
+
+{% block title %}Delete account{% endblock title %}
+
+{% block content %}
+
+<form action="" method="post">{% csrf_token %}
+    <p>Are you sure you want to delete your account?</p>
+    <p><strong>This operation cannot be undone. Your fills will be deleted.</strong></p>
```
jonc125, line 10:
```suggestion
    <p><strong>This operation cannot be undone. All entities owned by you (models, datasets, experiment results, etc) will be deleted.</strong></p>
```

jonc125 (approved):

PR #379, opened 2021-12-07
Update register.html

MauriceHendrix:
Updated the forwarding for already logged in users to go to home rather than to /, which works better in a subfolder

also updated the success_url in the view to use home via the url system rather than compiling it from / and the force_script_name


MauriceHendrix:
Updated the forwarding for already logged in users to go to home rather than to /, which works better in a subfolder

also updated the success_url in the view to use home via the url system rather than compiling it from / and the force_script_name

jonc125 (approved):

PR #410, opened 2022-04-11
Updated the graph legend

MauriceHendrix:
@mirams suggested a change to the graph legends.
**Before the change:**
image

**Now it looks like:**
image

(these graphs aren't exactly 100% the same, but it shows the principle)

### DIFF
--- a/weblab/static/js/visualizers/displayPlotFlot/displayPlotFlot.js
+++ b/weblab/static/js/visualizers/displayPlotFlot/displayPlotFlot.js
@@ -1,65 +1,6 @@
 var utils = require('../../lib/utils.js');
 var common = require('../../expt_common.js');
 
-/* toggle functionality to force groups on/off */
-function groupToggle(){
-    groupClass = $(this).attr('id').replace('selectGroup-', '.')
-    prefix = $(this).attr('id').replace('selectGroup-group-', '').split('-')[0];
-    totalNumChoices = $(`#${prefix}choices > .flotLegendEntity`).length;
-
-    togglerChecked = $(this).is(':checked');
-    groupCheckboxes = $(groupClass);
-
-    // set all checboxes on/off as appropriate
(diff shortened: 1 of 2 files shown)


MichaelClerx:
How did this get to 37 commits? And what is the change? I can see the difference, but what's the reasoning?

MauriceHendrix:
> How did this get to 37 commits? And what is the change? I can see the difference, but what's the reasoning?

it should be 1 commit only 2 files changed, not wure what's happened there!

MichaelClerx (approved):
It looks OK but I don't really know :D

MichaelClerx:
@MauriceHendrix does this address #412 and/or #413 ?

MauriceHendrix:
@MichaelClerx I don't think it does I think it's just a layout update.

MauriceHendrix:
> The 'Reset Zoom' and 'Toggle Legend' probably want to stay as the first thing next to the graph (with loads of groups the toggle legend button might disappear to where it is not clickable unless you have already toggled the legend!)

Really? I thought that would be confusing since those buttons relate to the individual models and not the groups?

mirams:
The 'Reset zoom' button is purely to do with the graph? I don't know exactly how the toggle legend button works, but I thought it was just "show all the legend entries (for groups, individual lines, etc.)" versus "cut it off after so many pixels" and I was worried that "so many pixels" might end up less than the size of the model groups area!

MauriceHendrix:
> The 'Reset zoom' button is purely to do with the graph? I don't know exactly how the toggle legend button works, but I thought it was just "show all the legend entries (for groups, individual lines, etc.)" versus "cut it off after so many pixels" and I was worried that "so many pixels" might end up less than the size of the model groups area!

true for the reset. the toggle works just on the individual models below it doesn't do anything with the groups above.

PR #416, opened 2022-06-21
Changed the way groups are selected and the editing is randered

MauriceHendrix:
main changes:
- fixes to toggles
- -ability to select multiple groups/models for a graph
-  ability to specify for which groups we want toggles

@agrawal-aditi @MichaelClerx  let me know if you're happy with this 

https://cardiac.nottingham.ac.uk/weblab/stories/79/edit


agrawal-aditi (approved):
lgtm

MichaelClerx (commented):
Any time I add or remove a model (or group) from the list, it forgets which protocol and CSV it was showing. This makes editing a bit difficult: I wanted to just add a model and now I have to undo that and go back to see which protocol+CSV it was!

MichaelClerx:
It would be a nice tweak if the "All" and all/any group options disappeared when there was only 1 model selected. (The `Select which groups can be switched on and off in the graph: ` objects can just be inactivated in this case, no need to remove the checkboxes etc!)

MauriceHendrix:
> Any time I add or remove a model (or group) from the list, it forgets which protocol and CSV it was showing. This makes editing a bit difficult: I wanted to just add a model and now I have to undo that and go back to see which protocol+CSV it was!

good point, just chenged the JS to re-apply the previously selected values (if and only if those remain possible options)

MauriceHendrix:
> It would be a nice tweak if the "All" and all/any group options disappeared when there was only 1 model selected. (The `Select which groups can be switched on and off in the graph: ` objects can just be inactivated in this case, no need to remove the checkboxes etc!)

@MichaelClerx I don't know which "All" and all/any group options do you mean? All options are names of groups? 
image

MichaelClerx:
Sorry, I meant in the graph! If I only have one model selected as options for the graph, the graph doesn't need a "toggle all", or any other group toggles for that matter

Issue indicated in red below:

Screenshot from 2022-06-28 17-59-44
Screenshot from 2022-06-28 17-59-49

Secondary issue in blue :D

MauriceHendrix:
@MichaelClerx  I think All is still useful

MauriceHendrix:
@MichaelClerx  I've done the requested changesI updated the CSS for the model selection changed the javascript, so that it fills in previously selected protocol etc when you change a model (only if those remain viable options). Also used js/css to not show the group toggles

MauriceHendrix:
@MichaelClerx  @agrawal-aditi  @agrawal-aditi  I've made the following changes, let me know if you think it's ready for roll-out on scrambler.

- updated the CSS for the model selection
- changed the javascript, so that it fills in previously selected protocol etc when you change a model (only if those remain viable options). 
- used js/css to not show the group toggles if you have only selected 1 group. (used CSS to "grey out" the label, so that it doesn't just disappear
- not showing "All" toggle if there is only 1 trace in the graph
-

agrawal-aditi:
I think the changes are fine.
One minor thing I observed is that the model selection pane has 'dropped down' (see image below).
Although, this returns to normal if I changes if I zoom out, pretty sure this was not happening before.
Nevertheless, I don't think it matters too much? 
selection_pane

MichaelClerx (approved):
Looks like it's working fine!

PR #427, opened 2022-07-26
Simpler graphs

MauriceHendrix:
- Simplified the way graphs for stories are set up (no longer using the comparisson code).
- still loading data in seperate files
- it should be much quicker to switch visualiser as the data does now not get re-downloaded when do


weblab/datasets/tests/test_views.py
```diff
@@ -465,7 +465,7 @@ def test_uses_octet_stream_for_unknown_file_type(self, mock_guess, my_dataset_wi
         )
 
         assert response.status_code == 200
-        assert response['Content-Type'] == 'application/octet-stream'
+        assert response['Content-Type'] == 'text/csv'
```
jonc125, line 468: Change the file extension on line 464 instead - this test is supposed to check a unrecognised file type!

weblab/datasets/views.py
```diff
@@ -204,7 +204,7 @@ def get(self, request, *args, **kwargs):
 
         content_type, _ = mimetypes.guess_type(filename)
         if content_type is None:
-            content_type = 'application/octet-stream'
+            content_type = 'text/csv'
```
jonc125, line 207: Not convinced by this! What files aren't being recognised as CSV and should be?

jonc125 (commented):
I've not checked the JS changes - easier to do by user testing I think - but thought I'd sanity-check the Django parts for you!

weblab/datasets/views.py
MauriceHendrix: without it, if I put the path to the CSV for building graphs in my browser (to check it out), the browser would not recognise it as CSV and try to download it as a file, instead of displaying it as nicely formatted CSV I could inspect.

jonc125: Can you delve into why this is, rather than pretending all unrecognised names are CSV? What is the filename that isn't being recognised? Is it a typical Web Lab pattern, in which case use https://docs.python.org/3/library/mimetypes.html#mimetypes.add_type to teach the library?

MauriceHendrix: seems to be a known bug with the python version we're running https://bugs.python.org/issue13952

MichaelClerx (approved):
I haven't checked the code, but the test server on cardiac seems to be doing really well!

PR #428, opened 2022-07-30
caching downloaidng experiment version files for graphs

MauriceHendrix:
Switched on caching for downloading teh files for graphs. Could possibly consider caching more things, but this is the main slowdown it sees and we're pretty sure that any changes => new version number so should be quite safe to do.


jonc125 (approved):
The Django side of this looks reasonable. I can't really assess the JS parts but if it's working in practice then happy to approve!

